PR #62, opened 2020-09-02
Make local mutations work against repc.

aboodman:


aboodman:
The test fixture changes just put it back the way it was prior to https://github.com/rocicorp/repc/commit/ddf72852bf0643b4b8b0be80f298c624605a43c1. I realized a different way to paper over the differences between replicache-client and repc.

src/replicache.test.ts
```diff
@@ -224,7 +224,9 @@ async function replicacheForTesting(
     diffServerAuth,
     diffServerURL,
     name,
-    repmInvoke: invoke,
```
arv, line 227: I intentionally used a function and not an object with invoke... Doesn't really matter since this code will go away at some point (or only be used by tests)... also more flexible.

src/repm-invoker.ts
```diff
@@ -3,6 +3,14 @@ import type {ScanItem} from './scan-item.js';
 import type {ScanOptions} from './scan-options.js';
 import type {DatabaseInfo} from './database-info.js';
 
+// TODO(repc-switchover): isWASM can go away, but so can this whole
+// type and all the machinery connected to it. Look at the commit
+// that introduced this to unwind it.
```
arv, line 8: I think we might want to keep it for replay tests?

src/scan-iterator.ts
```diff
@@ -130,27 +133,22 @@ export class ScanIterator<V> implements AsyncIterableIterator<V> {
...
-      opts: {
-        prefix: this._prefix,
-        start,
-        limit: scanPageSize,
-      },
-    });
+    };
+    const args = {
+      transactionId: this._transaction.id,
+      ...(this._isWASM ? {opts} : opts),
```
arv, line 143: or 

```js
opts: this._isWASM ? opts : undefined,
```

arv (commented):
LGTM

PR #66, opened 2020-09-05
Add a scan benchmark

aboodman:
Running benchmarks with REPMWASMInvoker please wait...
populate 1024x1000 (serial) | Median: 5890 | Throughput: 169.78 KB/s
populate 1024x1000 (parallel) | Median: 5271 | Throughput: 189.72 KB/s
scan 1024x1000 | Median: 1348 | Throughput: 741.84 KB/s
Done!
Running benchmarks with REPMHTTPInvoker please wait...
populate 1024x1000 (serial) | Median: 2374 | Throughput: 421.23 KB/s
populate 1024x1000 (parallel) | Median: 1743 | Throughput: 573.72 KB/s
scan 1024x1000 | Median: 666 | Throughput: 1.47 MB/s
Done!

### DIFF
--- a/sample/perf/index.html
+++ b/sample/perf/index.html
@@ -2,14 +2,11 @@
 <script type='module'>
   import Replicache, { REPMWASMInvoker, REPMHTTPInvoker } from './node_modules/replicache/out/mod.js';
 
-  const repmInvoker = location.hash == '#go' ?
-    new REPMHTTPInvoker('http://localhost:7002/') :
-    new REPMWASMInvoker();
-
-  console.log('Using repmInvoker', repmInvoker);
+  window.httpInvoker = new REPMHTTPInvoker('http://localhost:7002/');
+  window.wasmInvoker = new REPMWASMInvoker();
 
   let counter = 0;
-  async function makeRep() {
+  async function makeRep(repmInvoker) {
     const name = 'bench' + (counter++);
     await Replicache.drop(name, { repmInvoke: repmInvoker.invoke });
     return new Replicache({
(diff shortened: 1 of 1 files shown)


aboodman:
Note that this is not "fair" at all in the sense that the Go code is running natively on the machine and the WASM is obviously not. However, it's a bar that we have available to compare to.

We also know the size of data customers are importing and scanning, so the absolute speed is useful information for us too.

sample/perf/index.html
```diff
@@ -2,14 +2,11 @@
 <script type='module'>
   import Replicache, { REPMWASMInvoker, REPMHTTPInvoker } from './node_modules/replicache/out/mod.js';
 
-  const repmInvoker = location.hash == '#go' ?
-    new REPMHTTPInvoker('http://localhost:7002/') :
-    new REPMWASMInvoker();
-
-  console.log('Using repmInvoker', repmInvoker);
+  window.httpInvoker = new REPMHTTPInvoker('http://localhost:7002/');
```
arv, line 5: window.x :'(

Happy to fix it later but it does not really matter

The solution would have been to add the event listeners using code instead of html attributes.

sample/perf/index.html
```diff
@@ -69,30 +79,39 @@
...
   function humanSize(bytes) {
-    if (bytes == 0) { return "0.00 B"; }
+    if (bytes == 0) { return '0.00 B'; }
     var e = Math.floor(Math.log(bytes) / Math.log(1024));
     return (bytes / Math.pow(1024, e)).toFixed(2) + ' ' + ' KMGTP'.charAt(e) + 'B';
   }
 
-  main();
+  function log() {
+    document.querySelector('#output').value += Array.prototype.join.call(arguments, ' | ') + '\n';
```
arv, line 108: Why u no rest parameters?

aboodman: Just hard to keep track of all the sugar in these languages.

arv: How is `Arrray.prototype.join.call(arguments)` easier to keep track of? I guess old habits die hard =P

aboodman: I think I have a deep seated memory that `arguments` is weird and nothing you'd expect will work with it.

arv: That is correct. arguments is never needed in ES6. We killed it!

PR #67, opened 2020-09-06
A series of minor fixes that establish a baselines scan performance of about 30MB/s.

aboodman:
We should wire this up to run continuously on bots. It's super super easy to regress things accidentally (as this log shows).


sample/perf/index.html
```diff
@@ -56,10 +56,12 @@
     await populate(rep, opts);
     bench.reset();
     await rep.query(async (tx) => {
-      let i = 0;
+      var count = 0;
```
arv, line 54: why var?

arv (commented):
LGTM

phritz (commented):
LGTM

sample/perf/index.html
aboodman: Because I sux. Looks like you or I already fixed this.

PR #101, opened 2020-09-18
Add pusher init to main script block.

aboodman:
I didn't realize that non-module script was visible to module script.


arv (commented):
LGTM

PR #105, opened 2020-09-19
Rework the overlapping commits test.

aboodman:
It was testing a behavior that repc doesn't exhibit. Instead attempt to test that writes have exclusive access to the cache.

Fixes https://github.com/rocicorp/repc/issues/181

FYI @phritz

### DIFF
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -419,7 +419,7 @@ test('subscribe with error', async () => {
 });
 
 test('overlapping writes', async () => {
-  async function dbwait(tx: ReadTransaction, dur: number) {
+  async function dbWait(tx: ReadTransaction, dur: number) {
     // Try to take setTimeout away from me???
     const t0 = Date.now();
     while (Date.now() - t0 > dur) {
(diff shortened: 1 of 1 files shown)


src/replicache.test.ts
```diff
@@ -436,35 +425,43 @@ test('subscribe with error', async () => {
...
 
-  // Finish A.
-  ar.resolve();
-  await resAFuture;
-  expect(await rep.get('k')).to.equal('a');
+  // writes wait on writes
+  rep = await replicacheForTesting('conflict');
+  const mut = rep.register(
+    'wait-then-return',
+    async (tx, {duration, ret}: any) => {
```
arv, line 445: You can do it!

src/replicache.test.ts
```diff
@@ -436,35 +425,43 @@ test('subscribe with error', async () => {
...
-  const mutB = rep.register('mutB', async (tx, v) => {
-    await tx.put('k', v);
-    await br.promise;
-  });
+test('overlapping writes', async () => {
+  async function dbwait(tx: ReadTransaction, dur: number) {
+    // Try to take setTimeout away from me???
+    const t0 = Date.now();
+    while (Date.now() - t0 > dur) {
+      await tx.get('foo');
```
arv, line 433: Maybe use timerwait in here too? Right now it does as many `get` as it can do in the given time.

arv (approved):

src/replicache.test.ts
arv: I did it for you

src/replicache.test.ts
```diff
@@ -450,35 +439,43 @@ test('subscribe with error', async () => {
...
+    }
+  }
 
-  // Start A and B at the same commit.
-  const resAFuture = mutA('a');
-  const resBFuture = mutB('b');
+  async function timerwait(dur: number) {
```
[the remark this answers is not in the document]
aboodman: Is there a heuristic on when you recommend to use await and when not?

src/replicache.test.ts
aboodman: Cannot use timerwait because the transaction auto-commits when control returns to event loop and no idb ops are pending.

Can just do a busy wait, that might make this more clear.

src/replicache.test.ts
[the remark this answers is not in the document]
arv: My rule is to convert non promise behavior to promise behavior in an as simple as possible normal function. Then build up the more advanced features on top of that using async functions

src/replicache.test.ts
aboodman: Yeah I just can't keep all the ts stuff straight w/o more time in it. Thanks.

src/replicache.test.ts
arv: busy wait might get optimized away? This seems fine but maybe add a comment

PR #126, opened 2020-09-30
Add instructions to build for production with rollup.

aboodman:


sample/lit-todo/README.md
```diff
@@ -6,5 +6,8 @@ Absurdy small, fully functioning local-first todo list using Replicache.
 
 ```
 npm install
-npx @web/dev-server --node-resolve --watch  --root-dir=../../ --app-index=sample/lit-todo/index.html
+rollup -c --watch &
+python -m SimpleHTTPServer
```
arv, line 10: This one does not serve wasm files with the correct mime type :'(

sample/lit-todo/rollup.config.js
```diff
@@ -0,0 +1,10 @@
+import {nodeResolve} from '@rollup/plugin-node-resolve';
+
+export default {
+  input: 'main.js',
+  output: {
+    dir: 'output',
+    format: 'cjs',
```
arv, line 7: output should be esm

sample/lit-todo/main.js
```diff
@@ -0,0 +1,193 @@
...
+  document.body.style.visibility = 'visible';
+});
+document.querySelector('mwc-fab').addEventListener('click', handleCreate);
+
+Pusher.logToConsole = true;
+new Pusher('8084fa6056631d43897d', {
+  cluster: 'us3',
+})
+  .subscribe('u-2')
+  .bind('poke', function (data) {
```
arv, line 191: [taste] use arrow functions

sample/lit-todo/main.js
```diff
@@ -0,0 +1,193 @@
...
+  // will send batches of mutations here for application when the
+  // network is available.
+  batchURL: 'https://replicache-sample-todo.now.sh/serve/replicache-batch',
+
+  // Auth token for your client view and batch endpoints, if any.
+  dataLayerAuth: '2',
+  pushDelay: 300,
+
+  repmInvoker: new REPMWasmInvoker(
+    './output/wasm/release/replicache_client_bg.wasm',
```
arv, line 38: Should probably use import.meta.url but whatevs

arv: Actually, do you really need this? Doesn't the default work with rollup?

arv (commented):
LGTM

sample/lit-todo/main.js
aboodman: Fixed in subsequent commit.

sample/lit-todo/main.js
aboodman: No, the wasm bundle really needs to be a separate resource from the big bundle, I think. And without this, the script doesn't know where to find the wasm file.

arv: The wasm file is relative to `repm-invoker.js` so as long as we keep the file structure that should still work. With rollup we end up with a singe js file (I think) but we should probably put the wasm files as direct siblings to repm-invoker and just use `./replicache_client.js`

aboodman: The problem is that lower-level code currently defaults the name of the wasm module to name of including js module, which in this case in 'main.js', so it looks for 'main.wasm'.

arv: Yes. `replicache_client.js` uses its current path and appends `_bg.wasm` and with rollup `replicache_client.js` gets folded into the main output bundle.

sample/lit-todo/rollup.config.js
aboodman: I'm not actually sure what difference this makes given there are no imports, but done.

PR #138, opened 2020-10-17
Some documentation updates:

aboodman:
- Use interface instead of type for all option types and exclude from export. This causes the __named_parameters syntax at usage site and excludes the separate (useless) struct from the docs which is nice.
- Improve deprecatedness of REPMWasmInvoker and friends.


arv (commented):
LGTM

Makes sense that it would only expand the docs when using destructuring.

PR #140, opened 2020-10-17
Roll repc to 0.10.1 and add dropIndex

aboodman:


src/transactions.ts
```diff
@@ -184,6 +194,14 @@ export class WriteTransactionImpl extends ReadTransactionImpl
     });
   }
 
+  async dropIndex(name: string): Promise<void> {
+    throwIfClosed(this);
+    await this._invoke('dropIndex', {
+      transactionId: this.id,
+      name: name,
```
arv, line 201: Just:

`name,`

arv (commented):
LGTM

src/transactions.ts
aboodman: This has since been fixed.

PR #200, opened 2020-11-12
Add something about errors

arv:


arv:
I'm going to close this and open an issue to document this in the API docs instead.

README.md
```diff
@@ -110,6 +110,10 @@ const handleCheckbox = async (id, e) => {
 - We recommend [enabling console persistence](https://stackoverflow.com/questions/5327955/how-to-make-google-chrome-javascript-console-persistent) while developing replicache-enabled apps to make debugging easier.
 - Remember that data changes can happen "underneath" you and cause `subscribe()` to re-fire at any time. These changes can come from the server or from a different tab. If your UI is not reactive (driven solely by the data model) you need to take extra steps to ensure the UI is in sync with the data.
 
+### Errors
+
+Critical errors that prevent Replicache from working are always thrown. We also use `console.error` / `console.warn` for errors that does not prevent Replicache from working but might temporary show confusing data until the next sync (this can happen if a _mutator_ is missing for example). Console errors/warnings should be take action on either by you or us.
```
phritz, line 115: - "Console errors/warnings should be take action on either by you or us." -- at a lower level we remove this ambiguity by [saying](https://github.com/rocicorp/replicache/blob/master/contributing.md#style-general) that the definition of an *error* is that something truly unexpected happened and a developer should go look and do something. We do not have warnings at a lower level because (as stated in the rationale) not-actionable expected stuff emitted as a warning is just confusing. you even say that it is confusing. we should strive to not be confusing. in my opinion the thing that happened is either truly unexpected and so an error or truly expected and should not be scary so should be emitted as an info. i guess we can do what we want at the js level but having different log policies at different levels potentially makes things confusing. the advantage of the repc logging policy is that it makes it **super clear** to people what they have to pay attention to: errors. suggest adopting this at a higher level as well, but of course up to you.

phritz, line 115: "Critical errors that prevent Replicache from working are always thrown." -- this feels a bit vague to me. what does "critical" mean? does my application have to stop or can i continue trying to call into replicache? maybe we can make it more specific, either saying which errors or types of errors are fatal vs which there is some hope of continuing.

phritz (commented):
Thank you. Some additional suggestions.

README.md
arv: I agree that we should not have different policies at different layers.

To me there are two distinct cases; `console.error` (not warn*) and throw.
- `console.error` does not prevent the app using Replicache from continuing to work but "something truly unexpected has happened and a developer should go look"
- Exception: Replicache cannot continue to work and your app is in a broken state unless you explicitly handle the exception and do something about it.

[*] We use warn in one place in repc and never in replicache-sdk-js
https://github.com/rocicorp/repc/blob/d02aedd3f42b83e51b35f300b19405519eb08ece/src/embed/connection.rs#L129

phritz: that sounds really clear to me!

PR #213, opened 2020-11-18
Update lit-todo to Replicache 3.0, finally.

aboodman:
Gave up trying to properly migrate this tiny sample and just taught sw to force-update clients on update. Good enough.


sample/lit-todo/main.js
```diff
@@ -55,20 +54,20 @@ rep.onSync = syncing => {
 
 rep.subscribe(
   async tx => {
-    return (await tx.scanAll({indexName: 'byOrder'})).map(([_, v]) => v);
+    return (await tx.scanAll({indexName: 'byOrder2'})).map(([_, v]) => v);
```
arv, line 57: Why order2?

aboodman: Potential API usability issue: we have defined `createIndex` to throw if the named idx already exists but with a different definition, and drop throws if asked to drop and index that *doesn't* exist, and there's no `get`/`has` for index.

PR #214, opened 2020-11-18
Name the Replicache instance.

aboodman:
This is an easy way to start fresh on the client. Because backend is nuking data on every schema change, it makes clients unable to sync because they are at a sequence number far ahead of server.

Really what we need to do is put a version number in the protocol and nuke local state when it changes but ... another day.


arv (commented):
LGTM

PR #220, opened 2020-11-19
Whoops. Fix broken sync on redo.replicache.dev.

aboodman:
I changed the user id for the demo on the backend and changed it on the frontend, but the bit for pusher-js was non-DRY.

😬


arv (commented):
LGTM

PR #221, opened 2020-11-20
Do one intial sync on load.

aboodman:
This stopped working a few commits back when I took out the timer sync.


sample/lit-todo/main.js
```diff
@@ -201,8 +199,10 @@ function update(newTodos) {
 }
 
 // FOUC
-window.addEventListener('load', () => {
+window.addEventListener('load', async () => {
   document.body.style.visibility = 'visible';
+  await initIndexes();
```
arv, line 204: Why wait so long to do sync?

PR #224, opened 2020-11-23
Give up on service worker for awhile.

aboodman:
It's just making everything harder :(


aboodman:
Also address @arv complaint about waiting so long to sync.

PR #319, opened 2021-03-22
New setup.md doc with differential sync!

aboodman:


aboodman:
@arv @phritz this is OK to review now. I need to put in all the graphics but the copy is good to go.

Please try and actually go through the steps and provide feedback in both an editorial and technical sense.

This is going to be our main entrypoint for people actually trying to use Replicache so let's make it awesome.

aboodman:
Landing this so that it's easier to read/review.

aboodman:
You can read this now here: https://github.com/rocicorp/replicache-sdk-js/blob/main/doc/setup.md

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+
+You don't need REST or GraphQL APIs when you use Replicache, either. Instead, Replicache continuously synchronizes with your server in the background using two special endpoints: `replicache-push` and `replicache-pull`.
+
+To learn more about the big picture of Replicache, see [the design document](TODO).
+
+We will be building this particular sample app using [Next.js](https://nextjs.org/), [Supabase](https://supabase.io/), and [Pusher.js](https://pusher.com/) because they are great and work nicely with Replicache. But you can use pretty much any frontend and backend stack you want — just apply these steps to your situation, or [contact us](https://replicache.dev/#contact) for help.
+
+## Prerequisites
+
+You only need [Node.js](https://nodejs.org/en/), version 10.13 or higher to get started.
```
arv, line 23: Why 10.13?

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+export default async (req, res) => {
+  const push = req.body;
+  console.log('Processing push', JSON.stringify(push, null, ''));
+
+  const t0 = Date.now();
+  try {
+    const db = await getDB();
+    await db.tx(async t => {
+      const {nextval: version} = await db.one("SELECT nextval('version')");
+      let lastMutationID = parseInt(
```
arv, line 380: Does this not return a number? I'm confused why a BIGINT would be returned as string?

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+        (
+          await db.oneOrNone(
+            'SELECT last_mutation_id FROM replicache_client WHERE id = $1',
+            push.clientID,
+          )
+        )?.last_mutation_id ?? '0',
+      );
+
+      if (!lastMutationID) {
```
arv, line 389: `=== 0` for clarity?

arv: Also add a comment why

doc/setup.md
aboodman: That's what Next.js requires. Do we have a requirement for a bigger number?

doc/setup.md
aboodman: probably because the lib was written before js had bigint.

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+    "from": "Fred",
+    "order": 2,
+    "content": "Taaaacos"
+  }
+}
+```
+
+(A real app would likely separate out the user entities, but this is good enough for our purposes.)
+
+**Note:** A quick word on IDs: Unlike with classic client/server apps, Replicache apps can't rely the server to assign unique IDs to data items. That's because the client is going to be working with the data long before it reaches the server, and the client and server need a way to refer to items consistently. Therefore, Replicache requires that clients assign IDs. Browsers have [cryptographically strong random sources](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) now, so there's no real downside to this. If you think this might block your usage of Replicache, [reach out](https://replicache.dev/#contact) — we'd like to learn more.
```
phritz, line 48: This part is a little fuzzy for me. Clearly not all ids, right? User ids and changes in the data layer that do not originate in clients have server-created ids. Maybe we want to say that the server can't assign unique ids to client-created items?

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+
+_Note:_ Careful readers may be wondering what happens with the order field during sync. Can multiple messages end up with the same order? Yes! But in this case, what the user likely wants is for their message to stay rougly at the same position in the stream, and using the client-specified order and sorting by that roughly achieves the desired result. If we wanted better control over this, we could use [fractional indexing](https://www.npmjs.com/package/fractional-indexing) but that's not necessary in this case.
+
+Restart the server and you should now be able to make changes!
+
+TODO: Movie
+
+Notice that even though we're not saving anything to the server yet, the mutations are saved locally across sessions, and it even works across tabs. Whee!
+
+## Step 5: Remote Database Setup
```
phritz, line 296: We're having a little bit of sprawl in the terminology we use to describe the data layer. At times we say 'datastore', 'database', or 'data layer'. Or 'backend'. I think we should maybe standardize on 'backend'?

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+        }
+
+        console.log('Processing mutation:', JSON.stringify(mutation, null, ''));
+
+        switch (mutation.name) {
+          case 'createMessage':
+            await createMessage(db, mutation.args, version);
+            break;
+          default:
+            throw new Error(`Unknown mutation: ${mutation.name}`);
```
phritz, line 421: This halts the protocol. Maybe we should skip it instead?

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+import {getDB} from '../../db.js';
+
+export default async (_, res) => {
+  const db = await getDB();
+  await db.task(async t => {
+    await t.none('DROP TABLE IF EXISTS message');
+    await t.none('DROP TABLE IF EXISTS replicache_client');
+    await t.none('DROP SEQUENCE IF EXISTS version');
+    // Stores chat messages
+    await t.none(`CREATE TABLE message (
```
phritz, line 343: If index on message version elided intentionally for clarity, ok. But if you want it there, it's not :)

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+};
+```
+
+Voila. We're now round-tripping browsers and devices!
+
+TODO: Video
+
+Also notice that if we go offline for awhile, make some changes, then come back online, the mutations get sent when possible.
+
+We don't have any conflicts in this simple data model, but Replicache makes it easy to reason about most conflicts. See the [Design Doc](TODO) for more details.
```
phritz, line 540: Is the TODO here to link to the section in the dd or to link there and improve the example? Because the dd isn't as clear as it could be. Think we want to show explicitly how this works, with code. Guessing you are thinking the same, just checking.

doc/setup.md
```diff
@@ -0,0 +1,630 @@
...
+
+The only thing left is to make it live — we obviously don't want the user to have to manually refresh to get new data 🙄.
+
+## Step 8: Poke
```
phritz, line 544: I think it is worth mentioning that one can do this manually, eg pull on a timer. I think they don't have any control over push though do they? Might be good to clarify.

phritz (commented):
Looks great! 
- I made some small edits that I assume are not controversial here: https://github.com/rocicorp/replicache/pull/340
- some questions I didn't know how to resolve myself are below.
- I'm going to follow up with a proposal for a "mental model" section or separate doc that is a short, top-down description that describes in a lot less detail than the designdoc how Replicache works, eliding implementation details of replicache that aren't important to the customer. For someone like me I like to see the step by step but also to have a top-down thing to orient me ahead of time, and the designdoc has too much detail. Anyway, I'll follow up with a separate proposal for that.

doc/setup.md
arv: It just felt very specific

doc/setup.md
arv: Do we really need bigint? Isn't float64 sufficient (`2**53 - 1` safe ints).

Anyway... ignore me. I'm just nitpicking

PR #333, opened 2021-03-24
add launch checklist

phritz:


launch-checklist.md
```diff
@@ -1 +1,20 @@
+# Replicache Launch Checklist
```
aboodman: Nice. Can you create a new `doc/` folder and put it there? I'm going to be putting all the doc there.

launch-checklist.md
```diff
@@ -1 +1,20 @@
...
 
+Before you launch with Replicache in your product, it's a good idea to double-check that you have correctly covered all the small details of integration. This list can help you determine if you might have missed a detail, or deferred and then forgotten about it.
+
+- **JS SDK**
+  - if you wish to version the schema of the mutations in `Push` and the `clientView` returned in `Pull`, version information is provided in `ReplicacheOptions.schemaVersion`
```
aboodman: Grammatical nit, but this sounds informational and not like a task the dev should do. Rephrase like: "consider adding a `schemaVersion` field to `ReplicacheOptions` proactively so that you don't have to have a special case for "no schemaVersion" later.

aboodman: Please capitalize these.

launch-checklist.md
```diff
@@ -1 +1,20 @@
...
+  - re-auth is handled for `Push` and `Pull` via `getPushAuth` and `getPullAuth`
```
aboodman: Can your session token expire during a user session? If Replicache gets a 401 during push or pull, it will ...

launch-checklist.md
```diff
@@ -1 +1,20 @@
...
+- **All endpoints**
+  - the user is authenticated and the `clientID` verified to belong to that user
```
aboodman: I don't understand what the task is here for the dev

launch-checklist.md
```diff
@@ -1 +1,20 @@
...
+- **Push endpoint**
+  - the `lastMutationID` for a client is updated transactionally along with the pushed mutations' effects
```
aboodman: We need to create a *Server Requirements* doc that details exactly what is needed from server at a minimum. Like for example, I believe that CosmosDB's "consistent prefix" setting is consistent enough for Replicache's purposes.

launch-checklist.md
```diff
@@ -1 +1,20 @@
...
+  - all mutations with `id`s less than the client's current `lastMutationID` are skipped
+  - all mutations with `id`s greater than the client's current `lastMutationID+1` are ignored
+  - if a temporary error is encountered that might be resolved on retry, it halts processing mutations
```
aboodman: I would phrase this more as "think about what you want the error handling policy to be" and link to a doc, either the push endpoint reference doc which discusses this or a dedicated doc on error handling.

launch-checklist.md
```diff
@@ -1 +1,20 @@
...
+  - if a permanent error is encountered such that the mutation will never be appiable, ignore that mutation and increment the `lastMutationID` as if it were applied. Failing to do this can deadlock the client if a mutation will always fail.
```
aboodman: sp: applicable

aboodman (commented):
love this

launch-checklist.md
```diff
@@ -1 +1,21 @@
+# Replicache Launch Checklist
 
+Before you launch with Replicache in your product, it's a good idea to double-check that you have correctly covered all the small details of integration. This list can help you determine if you might have missed a detail, or deferred and then forgotten about one.
+
+- **JS SDK**
+  - if you wish to change the type of a mutator (eg, the number or type of its arguments) you must choose a new name; Replicache does not handle mutator versioning
+  - if you wish to version the schema of the mutations in `Push` and the `clientView` returned in `Pull`, version information is included in `ReplicacheOptions.schemaVersion`
```
arv:
```suggestion
  - if you wish to version the schema of the mutations in **Push** and the **clientView** returned in **Pull**, version information is included in `ReplicacheOptions.schemaVersion`
```

Push, clientView and Pull are not methods. We do have a `pull` method but that is it.

launch-checklist.md
```diff
@@ -1 +1,21 @@
...
+  - be sure that re-auth is handled for `Push` and `Pull` via `getPushAuth` and `getPullAuth`
```
arv:
```suggestion
  - be sure that re-auth is handled for **Push** and **Pull** via `getPushAuth` and `getPullAuth`
```

launch-checklist.md
```diff
@@ -1 +1,21 @@
...
+- **All endpoints**
+  - ensure the user is authenticated and the `clientID` verified to belong to that user
+- **Push endpoint**
+  - ensure that the `lastMutationID` for a client is updated transactionally along with the pushed mutations' effects
+  - all mutations with `id`s less than the client's current `lastMutationID` must be skipped
+  - all mutations with `id`s greater than the client's current `lastMutationID+1` must be ignored
```
arv: Maybe be more specific than skipped and ignored. Maybe explain why skipped/ignored.

Also, to me as a non english speaker, skipped and ignored are too subtle (I don't know the difference)

launch-checklist.md
```diff
@@ -1 +1,21 @@
...
+  - if a temporary error is encountered that might be resolved on retry, the endpoint must halt processing mutations and return
+  - if a permanent error is encountered such that the mutation will never be appiable, the endpoint must ignore that mutation and increment the `lastMutationID` as if it were applied. Failing to do this will deadlock the client if a mutation will always fail.
```
arv:
```suggestion
  - if a permanent error is encountered such that the mutation will never be appliable, the endpoint must ignore that mutation and increment the `lastMutationID` as if it were applied. Failing to do this will deadlock the client if a mutation will always fail.
```

launch-checklist.md
```diff
@@ -1 +1,21 @@
...
+  - ignore all `PushRequest`s with an unexpected `version`
```
arv:
```suggestion
  - ignore all **push requests** with an unexpected `version`
```

launch-checklist.md
```diff
@@ -1 +1,21 @@
...
+- **Pull endpoint**
+  - ensure that the `lastMutationID` returned in the response is read in the same transaction as the client view data (ie, is consistent with it)
+  - if there is a problem with the `cookie` (e.g., it is unusable) return all data
+  - ignore all `PullRequest`s with an unexpected `version`
```
arv:
```suggestion
  - ignore all **pull requests** with an unexpected `version`
```

arv (approved):
Just some minor things.

Most of them are that we should only use `foo` when we refer to something in the API (either the JS or the JSON format)

phritz:
Thanks this is great feedback. I believe I have addressed it all except in several cases on which I'll comment directly.

launch-checklist.md
phritz: Agree we should create such a doc. In the meantime, I've included the content inline.

phritz: Also PTAL at the "It is extremely important" item re consistency.

launch-checklist.md
phritz: ok, i was using code style for everything that is a technical entity, believe now code style is only for actual things we have in the code.

PR #343, opened 2021-03-27
Update setup.md

vitaly-t:
Your `pg-promise` usage example was completely invalid...

* Your import was wrong
* You re-initialize the whole library on each call, which is a horrible-horrible approach to using the library.
* You're leaking the database connections on every call

This change rectifies that.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -340,9 +340,7 @@ import * as pgInit from 'pg-promise';
 
 const pgp = pgInit({/* initialization options */});
 
-const db = pgp(process.env.REPLICHAT_DB_CONNECTION_STRING);
-
-export db;
+export const db = pgp(process.env.REPLICHAT_DB_CONNECTION_STRING);
 ```
 
 And another new file `pages/api/init.js` that initializes the schema:
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -335,26 +335,22 @@ npm install pg-promise
 
 Create a new file `db.js` with this code:
 
-```js
-import pgInit from 'pg-promise';
+```ts
+import * as pgInit from 'pg-promise';
 
-export async function getDB() {
-  const pgp = pgInit();
-  const db = pgp({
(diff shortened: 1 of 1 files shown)


doc/setup.md
```diff
@@ -335,26 +335,20 @@ npm install pg-promise
 
 Create a new file `db.js` with this code:
 
-```js
-import pgInit from 'pg-promise';
+```ts
+import * as pgInit from 'pg-promise';
```
aboodman, line 339: From my reading of https://nodejs.org/api/esm.html#esm_commonjs_namespaces it seems like the way it is now is correct:

> CommonJS modules consist of a module.exports object which can be of any type.
>
> When importing a CommonJS module, it can be reliably imported using the ES module default import or its corresponding sugar syntax:
> ```import { default as cjs } from 'cjs';
>
> // The following import statement is "syntax sugar" (equivalent > but sweeter)
> // for `{ default as cjsSugar }` in the above import statement:
> import cjsSugar from 'cjs';
> ```

What you propose seems like it should *not* work:
> the ECMAScript Module Namespace representation of a 
> CommonJS module is always a namespace with a default export 
> key pointing to the CommonJS module.exports value.
> 
> This Module Namespace Exotic Object can be directly observed
> either when using import * as m from 'cjs' or a dynamic import:
>
> ```
> import * as m from 'cjs';
> console.log(m);
> console.log(m === await import('cjs'));
> ```

I tested, and actually both ways seem to work. What am I missing?

aboodman (commented):
Thanks for the PR and feedback, and for `pg-promise`.

Note that this is running on Next.js which is serverless.  I was re-initializing the library on every call to `getDB`, but that call only happens once per request. So I was expecting that the initialization is only happening once per js context, just as your docs suggest.

But, I do see that this change causes a warning to go away under `next dev`. So the js context must be shared across requests at least in that case.

> Your import was wrong

I'm not seeing how it was wrong. See comment inline.

PR #346, opened 2021-03-30
Handle storage events correctly with memstore and other edge cases.

### LINKED ISSUE #345: Replidraw chugs with multiple tabs open
A customer reported this. Specifically drag performance is terrible.

I initially assumed it was #310, but still happens with #311.

aboodman:
Fixes #345.

### DIFF
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -258,15 +258,16 @@ export class Replicache implements ReadTransaction {
     });
     if (hasBroadcastChannel) {
       this._broadcastChannel = new BroadcastChannel(storageKeyName(this._name));
-      this._broadcastChannel.onmessage = e => {
-        this._checkChange(e.data);
-      };
+      this._broadcastChannel.onmessage = () => this._onStorage();
     } else {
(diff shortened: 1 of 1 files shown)


src/replicache.ts
```diff
@@ -358,10 +359,11 @@ export class Replicache implements ReadTransaction {
...
   // changes the db.
-  private _onStorage = (e: StorageEvent): void => {
-    if (e.key === storageKeyName(this._name)) {
-      this._checkChange(e.newValue as string);
-    }
+  private async _onStorage() {
+    // Cannot just use the value from the other tab, because it can be behind us.
+    // Also, in the case of memstore, it will have a totally different, unrelated
+    // hash chain.
+    this._checkChange(await this._getRoot());
```
aboodman, line 366: It's tempting to simply always re-read `this._getRoot()`, but I think Erik is correct to skip that in cases like a transaction just landing where we literally just received that info from the previous RPC.

arv:
LGTM

PR #349, opened 2021-04-02
Rework benchmarks to cleanup and add idb tests.

aboodman:


aboodman:
Sorry for gigantor review. I wanted to be very sure this was measuring the right thing, so I cleaned up the code too.

aboodman:
Note to self: This PR can't be landed yet because it changes the output format of the benchmark text which will break our continuous perf tests. It also doesn't yet run on its own website on vercel.

PR #362, opened 2021-04-20
Update license to allow up to 500 commercial clients and to restrict

aboodman:
competing uses.


aboodman:
FYI @arv @phritz

licenses/BSL.txt
```diff
@@ -7,16 +7,22 @@ Licensor:             Rocicorp, LLC
...
-                      instances for any purpose.
+Additional Use Grant: You may make use of the Licensed Work, provided you
+                      do not use it in any of the following ways:
+
+                      * Create a synchronization engine. A "synchronization
+                     engine" is a commercial or non-commercial offering that
+                     synchronizes data between two or more computers.
+
+                     * Use it for commercial purposes with more than 500
+                     monthly active client instances.
```
phritz, line 18: Do we want to switch the language to talk about "profile instances" or similar or should we just stick with "client" because it is pretty clear?

licenses/BSL.txt
```diff
@@ -7,16 +7,22 @@ Licensor:             Rocicorp, LLC
...
 
 Change Date:          2023-04-12
 
 Change License:       Apache License, Version 2.0
 
-For information about alternative licensing arrangements for the Software,
-please visit: https://replicache.dev/
+In order to use Replicache for commercial purposes with more than 500 monthly
+active client instances, see: https://replicache.dev/#price.
```
phritz, line 25: I know there is a contact section just below https://replicache.dev/#price but i think we should have a sentence in the price section that tells them to email us to sign up for a commercial plan.

phritz (commented):
Exciting to be coming back to the point where people might start paying us :)

aboodman:
I do want to switch to profile instances. But we need to do a lot of
documentation changes concurrently, so let's do it as part of dd?

licenses/BSL.txt
aboodman: Let's leave it as client for now as it is consistent with everything else.

licenses/BSL.txt
aboodman: I don't like having something so salesy in the license. Skipping this.

PR #399, opened 2021-06-08
Fix the comment for the unauthorized header in code comment

KeKs0r:
Since the code is specifically checking for the status code `401`.


arv:
Nothing left to merge. Thanks for the fix.

PR #434, opened 2021-07-13
Test ec runner

arv:


arv:
I'm leaving this as is for now.

It is not clear if starting/stopping the instance is worth the complexity and AWS is too cryptic for me to be able to reason about this.

Let's wait and see how much the EC2 instance ends up costing per month before making this call.

PR #467, opened 2021-09-01
Fix: compare to current main commit when computing changes from pull.

### LINKED ISSUE #466: Subscription not fired when optimistic mutation rejected by server
Steps to reproduce:

1. Modify replicache-sample-chat backend to increment lmid on mutations, but not do anything. Leave optimistic mutations in place.
2. Add a todo
3. Wait for pull

Expected: newly added todo should disappear on pull (because server has acknowledged change, but server state does not include the todo).

Actual: todo remains until refresh or some other action that re-queries the data in Replicache, then disappears

aboodman:
When computing changed keys for a pull, we were comparing the last snapshot on main to the current state of the sync branch.
    
This is not correct, because main will typically have subsequent pending changes which have not been confirmed by server yet. We need to consider these changes too when computing the diff.
    
Fixes #466

### DIFF
--- a/src/sync/pull.ts
+++ b/src/sync/pull.ts
@@ -259,24 +259,14 @@ export async function maybeEndTryPull(
     // TODO check invariants
 
     // Compute diffs (changed keys) for value map and index maps.
-    const mainSnapshotMap = await prolly.Map.load(
-      mainSnapshot.valueHash(),
-      dagRead,
-    );
+    const mainHead = await db.Commit.fromHash(mainHeadHash, dagRead);
+    const mainHeadMap = await prolly.Map.load(mainHead.valueHash(), dagRead);
     const syncHeadMap = await prolly.Map.load(syncHead.valueHash(), dagRead);
-    const valueChangedKeys = prolly.Map.changedKeys(
-      mainSnapshotMap,
-      syncHeadMap,
-    );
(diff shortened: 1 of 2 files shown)


src/sync/pull.ts
```diff
@@ -259,24 +259,14 @@ export async function maybeEndTryPull(
     // TODO check invariants
 
     // Compute diffs (changed keys) for value map and index maps.
-    const mainSnapshotMap = await prolly.Map.load(
-      mainSnapshot.valueHash(),
-      dagRead,
-    );
+    const mainHead = await db.Commit.fromHash(mainHeadHash, dagRead);
+    const mainHeadMap = await prolly.Map.load(mainHead.valueHash(), dagRead);
     const syncHeadMap = await prolly.Map.load(syncHead.valueHash(), dagRead);
```
arv, line 264: Promise.all

arv: nm. I'll do it

arv: 8a991ca8e3c187792e3ae10cff326fabf89ecd1c

arv: 8a991ca8e3c187792e3ae10cff326fabf89ecd1c

arv:
I'm not fully sure why this is the case? I knew I was right to ask for detailed review on exactly this part when I first wrote this code since this is not clear to me.

PR #471, opened 2021-09-02
git grep -i wasm | 🔪

aboodman:


src/replicache.ts
```diff
@@ -1055,21 +1055,6 @@ export class Replicache<MD extends MutatorDefs = {}>
...
-   * When this is set to `true` the internal Replicache wasm module will log
-   * more things to the console (using `console.debug`). Setting this to false
-   * reduces the amount of logging done by the wasm module.
-   *
-   * If you want to see the verbose logging from Replicache in Devtools/Web
-   * Inspector you also need to change the console log level to `Verbose`.
-   * @deprecated
-   */
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  async setVerboseWasmLogging(): Promise<void> {
```
arv, line 1070: This is a breaking change.

Did we really want to do a major semver?

aboodman:
On Thu, Sep 2, 2021 at 11:54 AM Erik Arvidsson ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/replicache.ts
> <https://github.com/rocicorp/replicache/pull/471#discussion_r701447617>:
>
> > @@ -1055,21 +1055,6 @@ export class Replicache<MD extends MutatorDefs = {}>
>
>      return {result, ref};
>    }
> -
> -  /* c8 ignore next 13 */
> -  /**
> -   * When this is set to `true` the internal Replicache wasm module will log
> -   * more things to the console (using `console.debug`). Setting this to false
> -   * reduces the amount of logging done by the wasm module.
> -   *
> -   * If you want to see the verbose logging from Replicache in Devtools/Web
> -   * Inspector you also need to change the console log level to `Verbose`.
> -   * @deprecated
> -   */
> -  // eslint-disable-next-line @typescript-eslint/no-unused-vars
> -  async setVerboseWasmLogging(): Promise<void> {
>
> This is a breaking change.
>
> Did we really want to do a major semver?
>
You just changed the hashes so I think we're already there, right?

arv:
Good point. I have a few more hash changing changes so I want them in before we do another major release...

PR #478, opened 2021-09-05
Fix bug introduced in 48c699c27abf62565360a771d4fb935f7ab8f95a:

aboodman:
All our sample apps directly mutate the data returned by get() (and less commonly scan()). This was breaking change detection completely.


package.json
```diff
@@ -59,7 +59,9 @@
     "out/replicache.dev.wasm"
   ],
   "dependencies": {
+    "@types/lodash-es": "^4.17.4",
```
aboodman, line 62: Pulling this in for `cloneDeep`. This adds about 5kb of code to the binary :-(, but writing a deep clone -- even if restricted to json types -- has to concern itself with cycle detection. I feel like that's more your wheelhouse than mine.

src/json.test.ts
```diff
@@ -54,3 +54,14 @@ test('JSON deep equal', () => {
 
   t({a: 1, b: 2}, {b: 2, a: 1});
 });
+
+test('JSON deepFreeze', async () => {
+  const obj: any = deepFreeze({obj:{}, arr:[]});
```
aboodman, line 59: This causes ts warnings that I can't figure out how to silence.

```
  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore
```

... doesn't work.

aboodman (commented):
`prolly.Map` is way over complex now that it's just a sorted map. There's no reason for the separate `_pending` but I wanted to fix it and add tests with minimal changes before attempting change.

package.json
arv: That seems fine for now.

src/json.test.ts
arv: I'll look

src/json.ts
```diff
@@ -9,13 +9,25 @@ export type JSONValue =
   | Array<JSONValue>
   | JSONObject;
 
+/** Like `JSONValue` but deeply immutable */
```
arv, line 12: The docs can link these if you write `[[JSONValue]]`

src/prolly/map.ts
```diff
@@ -1,104 +1,131 @@
...
-    switch (p) {
-      case deleteSentinel:
-        return undefined;
-      case undefined:
-        return this._baseGet(key);
-      default:
-        return p;
-    }
+  has(key: string): boolean {
+    const {found} = this._binarySearch(key);
```
arv, line 83: inline

src/prolly/map.ts
```diff
@@ -1,104 +1,131 @@
...
-      return undefined;
-    }
-    const {found, index} = this._base.binarySearch(key);
+  get(key: string): ImmutableJSONValue | undefined {
+    const {found, index} = this._binarySearch(key);
     if (!found) {
       return undefined;
     }
-    return this._base.entries[index][1];
+    const [, value] = this._entries[index];
```
arv, line 92: inline

src/prolly/map.ts
```diff
@@ -1,104 +1,131 @@
...
+    const {found, index} = this._binarySearch(key);
+    if (found) {
+      this._entries.splice(index, 1);
+    }
+    this._changedKeys.add(key);
   }
 
   entries(): IterableIterator<Entry> {
-    return new Iter(this._base, this._pending);
+    return this._entries[Symbol.iterator]();
```
arv, line 116: or jusr `.values()`

src/prolly/map.ts
```diff
@@ -1,104 +1,131 @@
...
     return this.entries();
   }
 
   async flush(write: dag.Write): Promise<string> {
-    const newBase = new Leaf([...this]);
-    await write.putChunk(newBase.chunk);
-    this._base = newBase;
-    this._pending.clear();
-    return this._base.chunk.hash;
+    // TODO: how come ts doesn't complain about passing the mutable array?
```
arv, line 124: I played around with this yesterday and it makes sense but it is not very helpful for us.

src/prolly/map.ts
```diff
@@ -1,104 +1,131 @@
...
+    const chunk = dag.Chunk.new(this._entries, []);
+    await write.putChunk(chunk);
+    this._entries = [...this._entries];
```
arv, line 127: do we need a new array here?

arv:
I like how this gets rid of the leaf iterator! But maybe that was not the intent of this PR

src/json.ts
aboodman: done

src/prolly/map.ts
aboodman: This isn't in the newer PR, but for my info, can you explain what you mean here? I don't understand your comment.

src/prolly/map.ts
aboodman: I filed https://github.com/rocicorp/replicache/issues/483.

src/json.test.ts
aboodman: I figured it out (actually just clicked the little hint from vs code, which figured it out for me).

aboodman:
> I like how this gets rid of the leaf iterator! But maybe that was not the intent of this PR

Yeah I would like to do it, but this PR got out of hand, so I narrowed it and moved that idea to: https://github.com/rocicorp/replicache/issues/481

src/prolly/map.ts
arv: I just wanted it to be: `return this._binarySearch(key).found`

BTW. I'm applying this change.

PR #479, opened 2021-09-05
Return deep copies of Replicache data out of read interfaces.

aboodman:
This is important because the common pattern in applications (and which we advocate) is to directly mutate the returned data and re-write it. If we are caching that data internally weird things happen.

This PR also adds some typesafety internally to try and prevent us from making similar mistakes, but it's incomplete. See follow ups:

rocicorp/replicache#480
rocicorp/mono#136

### DIFF
--- a/src/deep-readonly.ts
+++ b/src/deep-readonly.ts
@@ -1,0 +1,15 @@
+export type DeepReadonly<T> = T extends (infer R)[]
+  ? DeepReadonlyArray<R>
+  : // eslint-disable-next-line @typescript-eslint/ban-types
+  T extends Function
+  ? T
+  : // eslint-disable-next-line @typescript-eslint/ban-types
+  T extends object
+  ? DeepReadonlyObject<T>
+  : T;
+
+export type DeepReadonlyArray<T> = ReadonlyArray<DeepReadonly<T>>;
+
+export type DeepReadonlyObject<T> = {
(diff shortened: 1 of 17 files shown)


src/deep-readonly.ts
```diff
@@ -0,0 +1,15 @@
+export type DeepReadonly<T> = T extends (infer R)[]
+  ? DeepReadonlyArray<R>
+  : // eslint-disable-next-line @typescript-eslint/ban-types
+  T extends Function
```
arv, line 4: we do not really need this case

src/json.ts
```diff
@@ -1,4 +1,6 @@
+import {cloneDeep} from 'lodash-es';
```
arv, line 1: you can probably do something like `from 'lodash-es/cloneDeep';`

src/kv/idb-store.ts
```diff
@@ -56,8 +57,13 @@ class ReadImpl {
     return (await wrap(objectStore(this._tx).count(key))) > 0;
   }
 
-  async get(key: string): Promise<Uint8Array | undefined> {
-    return wrap(objectStore(this._tx).get(key));
+  async get(key: string): Promise<Value | undefined> {
+    const v = await wrap(objectStore(this._tx).get(key));
+    if (v instanceof Uint8Array) {
```
arv, line 62: In theory we should probably freeze this too.

arv (commented):
LGTM

PR #492, opened 2021-09-07
Add utility to open a window and run code in it for tests.

whiten:
Use this to start a test for navigator.locks, verifying that it works on
all browsers we care about.

When test-util.ts newTab() is called, a new browser window is opened,
with all console logs from it funnelled back to the main test process. A
module (usually that of the current test) may be loaded and all exported
symbols made available, allowing type-checked Typescript code to be
executed easily. The run() method on the resulting tab may then be
called to execute snippets of code, allowing for testing of interaction
between two windows at the same origin.

Example:

```
import {newTab} from './test-util';

test('storage', async () => {
  const tab = await newPage();
  await tab.run('window.localStorage.setItem("test", "modified")');
  expect(window.localStorage.getItem('test')).to.equal('modified');
}
```


src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
   }
   return utf8.encode(s);
 }
+
+export function defineTestFunctions(self: Window & typeof globalThis): void {
+  // Mocha functions aren't available when importing from a new page.
+  if (typeof self.suiteSetup !== 'function') {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    const noop = function () {};
```
arv: [fyi] Another way to disable this test is to add a comment (I believe)

```
const noop = () => {
  // intentionally empty
};
```

arv: Prefer arrow functions over function expressions. FunctionDeclaration is preferred at top level (and where required by semantics)

arv: You can also do `const noop = () => undefined` which would have the same semantics as an empty function

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+    self.test = test as unknown as Mocha.TestFunction;
+  }
+}
+
+defineTestFunctions(self);
+
+type Response = {
+  tabId: string;
+  id: string;
+  result?: unknown;
```
arv, line 66: optional unknown seems a bit strange. Just do unknown?

There is a new feature in TS4.4 which allows it to  distinguish between missing properties vs `| undefined`. We are not using that yet.

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+};
+
+export async function initializeNewTab(event: MessageEvent): Promise<void> {
+  const {tabId, path} = event.data;
+
+  ['error', 'warn', 'info', 'debug', 'log'].forEach(level => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore No index signature of type 'string' was found on 'Console'.
+    console[level] = (...data: unknown[]) => {
+      window.opener.postMessage(
```
arv: You can probably just call this synchronously across the frame barrier.

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+        {tabId, id: 'log', result: {level, message: data.join(' ')}},
```
arv: just `message: data` seems more accurate (and flexible)

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+        window.origin,
+      );
+    };
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
+  let module: typeof import('./test-util');
```
arv, line 93: Interesting. I never had to import myself but I can see why you want this here.

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+  if (path !== undefined) {
+    try {
+      module = await import(window.origin + '/' + path);
```
arv: You probably should do something like:

```ts
new URL('/' + path, location.href).toString();
```

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+    } catch (e) {
+      console.log('Error importing ' + path + ':', e);
+    }
+  }
+
+  const handle = async (event: MessageEvent) => {
+    if (event.source === null || event.source != window.opener) {
```
arv: Prefer triple equals at all times

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+      return;
+    }
+    const {id, expr} = event.data;
+    const response: Response = {tabId, id};
+    if (expr) {
+      try {
+        response.result = await AsyncFunction(
```
arv: I see what you are doing here but I think you can make this simpler. Picking up the `AsyncFunction` function should not be needed. You can do this with plain old `Function`.

```ts
Function('module', `
  "use strict";
  Object.assign(self, module);
  return (async () => {
    ${expr}
  })();
`)(module);
```

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+  >
+> = {};
+
+function listener(event: MessageEvent<Response>): void {
+  const {tabId, id, result, error} = event.data;
+  const tabCallbacks = callbacks[tabId];
+  if (tabCallbacks === undefined) {
+    return;
+  }
+  if (id == 'log') {
```
arv: I think this would be simpler with direct calls instead of postMessage.

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+
+  if (!Object.keys(callbacks).length) {
+    window.addEventListener('message', listener, false);
+  }
+  callbacks[tabId] = {init: {resolve}};
+  const tabCallbacks = callbacks[tabId];
+
+  // Open a new tab to a URL that doesn't run anything on load.
+  const tab = window.open(
+    document.location.origin + '/src/test-util.ts',
```
arv: This loads the ts file as text. Probably not very meaningful. You can open `''` and you have communicate between these windows syncronously.

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+      let testutil = await import(window.origin + '/src/test-util.ts');
+      testutil.initializeNewTab(event);
+      window.removeEventListener('message', initialize, false);
+    };
+    window.addEventListener("message", initialize, false);
+  `;
+  tab.document.head.appendChild(script);
+
+  // Start initialization process, waiting for page to become live.
+  const interval = setInterval(() => {
```
arv, line 200: The `script` element has an onload event. Not sure if that helps here though?

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+    tab.postMessage({id: 'init', tabId, path: filename}, window.origin);
+  }, 100);
+  await promise;
+  clearInterval(interval);
```
arv: Love this pattern!

src/test-util.ts
```diff
@@ -32,3 +34,195 @@ export function b(
...
+
+  return {
+    run: async (expr: string) => {
+      const id = uuid();
+      const {promise, resolve, reject} = resolver<unknown>();
+      tabCallbacks[id] = {resolve, reject};
```
arv: I would use a Map for a map

src/web-lock.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      name,
+      {ifAvailable: true, mode: 'shared'},
+      async lock => lock !== null,
+    )) || false
+  );
+}
+
+test('ifAvailable', async () => {
+  await navigator.locks.request('exlock', {mode: 'shared'}, async () => {
+    expect(await tab.run('await lockIfAvailable("exlock")')).to.equal(true);
```
arv, line 38: I love how the exported bindings are also loaded into the new tab. Awesome work!

arv (changes requested):
This is great!

I'm extra nitpicky since this is your first PR.

I think the main thing I would change is to make the communcation sync and not rely on postMessage. It would simplify a lot of the code I believe

For sync communications you can do:

```js
const w = window.open();
w.onload = () => {
  console.log(w.document.body.textContent);
  console.assert(w.opener === window);
};
```

Remember that all global (`globalThis`/`window`) bindings (variabels, functions etc) are visible. So if you set `globalThis.foo` in the opened window the opener can call this directly (it can be an async function etc).

aboodman:
One thing to think about here is the fact that it's possible for different tabs from the same origin to be in separate processes, with their own separate main threads. In the more common case, every tab from the same origin will be in the same process and on the same thread / message loop. It's worth asking whether we want to test that case.

The way you are opening the window here, the two tabs will be in same process and you can in fact do cross-frame synchronous calls as Erik mentions. If you want to cause the opened window to have its own process, then you have to open it with `rel=noopener`, but in that case the synchronous calls will stop working.

See https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/ for more.

arv:
@aboodman Good point. Let's keep the postMessage goop for that reason.

src/test-util.ts
whiten: If I omit the `?`, I get the below error later in the file. Anything else to try here?

```
Property 'result' is missing in type '{ tabId: any; id: any; }' but required in type 'Response'.`
test-util.ts[66, 3]: 'result' is declared here.
```

src/test-util.ts
whiten: I played with this a bunch, and was unable to prove to myself that it ever fired. I'm not confident it wouldn't work, but am giving up for now.

src/test-util.ts
whiten: Thanks for the `undefined` tip, I didn't know that! Searched for other non-arrow functions and didn't find any.

src/test-util.ts
whiten: Yep, that's right. I had struggled to figure out how to then pass the resulting array back to message.log() and friends, but just got `...data` working. Thanks!

src/test-util.ts
whiten: It's the same domain though, allowing the use of `localStorage`! I didn't want to use the root page on this domain because that auto-runs the tests.

src/test-util.ts
whiten: Thanks!

src/test-util.ts
whiten: Done, good suggestion.

src/web-lock.test.ts
whiten: Thank you! I went through several drafts where I had different snippets of pure JS in quoted strings, and it was just proving impossible to debug. I'm glad I figured out how to make this work.

src/test-util.ts
whiten: Fixed here and three other places in this PR. I'm going to follow up with an add of `eqeqeq` to .eslintrc.cjs to help avoid this.

src/test-util.ts
whiten: Good call, done. With an await on the resulting Function, I don't need to do one inside.

whiten:
Okay, responded to all suggestions, sticking with `postMessage` to exercise cross-process cases as well. I accidentally introduced a merge commit into this, so let me know if you need me to rebase this again to excise that from the commit history here. Back to you, @arv!

src/test-util.ts
arv: haha

OK TypeScript, you win this time.

src/test-util.ts
arv: `''` is special

src/test-util.ts
```diff
@@ -72,91 +69,103 @@ type Response = {
...
+          `"use strict";
+           Object.assign(self, module);
+           return (async () => { ${expr} })();`,
         )(module);
       } catch (e) {
         response.error = e;
       }
     }
-    event.source.postMessage(response);
+    opener.postMessage(response, origin);
```
arv: `opener` will not work if out of process I believe?

arv: The old code should work on the other hand

src/test-util.ts
```diff
@@ -72,91 +69,103 @@ type Response = {
...
   };
 
-  window.addEventListener('message', handle, false);
+  addEventListener('message', handle, false);
   void handle(event);
 }
 
+function withTimeout<T = void>(promise: Promise<T>, ms: number) {
+  const timeout = new Promise((_resolve, reject) => {
+    setTimeout(() => reject(new Error(`Timeout after ${ms}ms.`)), ms);
```
arv, line 131: you could clearTimeout when the other one is done but this is just test code so no big deal

src/test-util.ts
```diff
@@ -189,39 +199,39 @@ export async function newTab(filename?: string): Promise<Tab> {
...
-      tabCallbacks[id] = {resolve, reject};
-      if (expr.search('return ') == -1) {
+      tabCallbacks.set(id, {resolve, reject});
+      if (expr.search('return ') === -1) {
         expr = 'return ' + expr;
       }
-      tab.postMessage({id, expr}, window.origin);
-      const result = await promise;
-      delete callbacks[id];
+      tab.postMessage({id, expr}, origin);
```
arv: I actually prefer `window.origin` over `origin`. I hardly every use `window.origin` so seeing it as identifier expression makes me look for the variable.

It is kind of a black art which properties of window are ok to access as identifier expressions so feel free to use your own judgement

arv (approved):

arv (approved):
Feel free to deal with `opener` when/if it becomes an issue

arv:
I see some test failures as well.

PR #494, opened 2021-09-08
Tighten up a few lint checks.

whiten:
Also tweak one prettier option to key property quoting consistent within
an object.

### DIFF
--- a/src/asserts.ts
+++ b/src/asserts.ts
@@ -27,8 +27,7 @@ export function assertArray(v: unknown): asserts v is unknown[] {
 
 export function invalidType(v: unknown, t: string): string {
   let s = 'Invalid type: ';
-  if (v == null) {
-    // includes undefined too because `==`
+  if (v === null || v === undefined) {
     s += v;
   } else {
     s += `${typeof v} \`${v}\``;
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -733,7 +733,7 @@ export class Replicache<MD extends MutatorDefs = {}>
             ? this.getAuth()
             : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
               this.getPushAuth!());
-          if (auth != null) {
+          if (auth !== null) {
             this.auth = auth;
             // Try again now instead of waiting for next push.
             return await this._invokePush(maxAuthTries - 1);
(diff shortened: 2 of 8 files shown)


src/replicache.ts
```diff
@@ -733,7 +733,7 @@ export class Replicache<MD extends MutatorDefs = {}>
             ? this.getAuth()
             : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
               this.getPushAuth!());
-          if (auth != null) {
+          if (auth !== null) {
```
whiten: Please be esp. careful in reviewing this change and the one at 803, as I'm not precisely sure what condition we mean to be testing for here. Thanks!

src/asserts.ts
```diff
@@ -27,8 +27,7 @@ export function assertArray(v: unknown): asserts v is unknown[] {
 
 export function invalidType(v: unknown, t: string): string {
   let s = 'Invalid type: ';
-  if (v == null) {
-    // includes undefined too because `==`
+  if (v === null || v === undefined) {
```
arv, line 30: Or disable eslint on this line

src/replicache.ts
arv: This needs to be `!= null` or `auth !== null && auth !== undefined` because the type def for `getAuth` allows returning both `null` and `undefined`.

src/sync/js-request.ts
```diff
@@ -10,7 +10,7 @@ export function callJSRequest<Body, Return>(
       // eslint-disable-next-line @typescript-eslint/naming-convention
       'Content-type': 'application/json',
       // eslint-disable-next-line @typescript-eslint/naming-convention
-      Authorization: auth,
+      'Authorization': auth,
```
arv, line 13: Like it

arv (changes requested):

src/replicache.ts
whiten: Will change to encompass both. That means that once `getAuth()` has returned a non-null/undefined value it can never change `this.auth` back to undefined, and I wasn't sure if that was the intention or not. :)

whiten: In this case, can it just be `if (auth) { ...`?

arv (approved):

PR #508, opened 2021-09-13
Rework newTab() utility to work with 'noopener'.

whiten:
This reimplements its internal message system to use localStorage events
rather than postMessage(). This does not mean that postMessage can't be
used for tests that pass {opener: true} to newTab, but makes the
underlying mechanism work regardless.


src/test-util.ts
```diff
@@ -72,21 +71,23 @@ type LogResponse = {
...
-  const {opener} = window;
-  if (opener === null) {
-    return;
-  }
+export async function initializeNewTab(
+  tabId: string,
+  path: string,
+  event: StorageEvent,
+): Promise<void> {
+  const send = (message: unknown) => {
```
arv: We have been using `JSONValue` for these but unknown is OK too.

src/testutil-new-tab.html
```diff
@@ -1,12 +1,17 @@
 <html>
   <head>
     <script type="application/javascript">
```
arv: Remove type attribute. JS is default

src/testutil-new-tab.html
```diff
@@ -1,12 +1,17 @@
 <html>
```
arv: I missed this file last time :-)

Add `<!doctype html>`

src/testutil-new-tab.html
arv: maybe even use `type="module"`which opts you into strict module mode.

src/web-lock.test.ts
```diff
@@ -4,8 +4,13 @@ import {newTab, Tab} from './test-util';
 
 let tab: Tab;
 
+function isFirefox(): boolean {
+  return navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
```
arv: Use a regxp?

```suggestion
  return /firefox/i.test(navigator.userAgent);
```

src/test-util.ts
arv: We have been using `JSONValue` for these but `unknown` is OK too.

arv (approved):

src/test-util.ts
whiten: Switched this to `JSONObject`, and used `JSONValue` in a few other places. Thanks!

src/testutil-new-tab.html
whiten: Thanks, done.

src/testutil-new-tab.html
whiten: Done.

src/web-lock.test.ts
whiten: Good call. I definitely didn't use StackOverflow for this. ;)

src/test-util.ts
whiten: Done.

src/testutil-new-tab.html
whiten: I also renamed this file to `src/test-util-new-tab.html` to match the punctuation on `src/test-util.ts`.

PR #509, opened 2021-09-14
docs: Fix typo, misspelled Subscribe

tmcw:
Does what it says on the tin, just noticed this typo while reading through the docs!


arv:
Thanks. I'll do this in different PR since i assume you do not care about signing a CLA just for this minor typo?

If you do want to sign a CLA see https://github.com/rocicorp/replicache/blob/main/contributing.md#legal

tmcw:
🤷 that's all right - just signed the CLA

PR #519, opened 2021-09-15
Add some benchmarks for navigator.locks.

whiten:
So far this looks too good to be true:

```
% npm run perf -- --lock
...
Running benchmarks please wait...
unheld read x 6486.70 ops/sec B11100.0% (3244 runs sampled)
held read x 6438.00 ops/sec B1350.0% (3219 runs sampled)
held read other tab x 6704.66 ops/sec B11400.0% (3353 runs sampled)
held read other process x 6663.33 ops/sec B1600.0% (3333 runs sampled)
unheld write acq x 6534.69 ops/sec B11700.0% (3268 runs sampled)
Done!
```


perf/lock.ts
```diff
@@ -0,0 +1,93 @@
...
+  return [
+    benchmarkUncontendedRead(),
+    benchmarkHeldRead(Mode.SameTab),
+    benchmarkHeldRead(Mode.OtherTab),
+    benchmarkHeldRead(Mode.OtherProcess),
+    benchmarkUncontendedWrite(),
+  ];
+}
+
+export async function isHeld(name: string) {
```
arv: I think eslint complains about missing return type on all exported functions

perf/lock.ts
```diff
@@ -0,0 +1,93 @@
...
+    {ifAvailable: true},
+    async lock => lock === null,
+  );
+}
+
+export function lockShared(name: string) {
+  const p = new Promise(res => {
+    window['resolve'] = res;
+  });
+  navigator.locks.request(name, {mode: 'shared'}, async () => p);
```
arv: useless `async`

perf/perf.js
```diff
@@ -9,6 +9,7 @@ import {
   benchmarkSubscribeSetup,
 } from './replicache.js';
 import {benchmarkIDBRead, benchmarkIDBWrite} from './idb.js';
+import {benchmarks as lockBenchmarks} from './lock.ts';
```
arv: if these are now running with the esbuild plugin we should remove all the extensions from the import paths for consistency.

perf/perf.js
```diff
@@ -138,7 +139,7 @@ const benchmarks = [
   benchmarkSubscribeSetup({count: 10}),
   benchmarkSubscribeSetup({count: 100}),
   benchmarkSubscribeSetup({count: 1000}),
-];
+].concat(lockBenchmarks());
```
arv:
```suggestion
  ...lockBenchmarks(),
];
```

perf/lock.ts
```diff
@@ -0,0 +1,93 @@
...
+  return await navigator.locks.request(
+    name,
+    {ifAvailable: true},
+    async lock => lock === null,
+  );
+}
+
+export function lockShared(name: string) {
+  const p = new Promise(res => {
+    window['resolve'] = res;
```
arv: maybe use a less generic name?

arv (approved):

perf/lock.ts
whiten: Semantically, yes, but Typescript insists. :)

perf/lock.ts
whiten: Yeah, in the first commit I didn't have the configs quite right. ESLint now flags all of these, and I fixed them. Thanks.

perf/perf.js
whiten: Done.

perf/perf.js
whiten: Done, thanks.

perf/lock.ts
whiten: Changed to use a map by lock name, since once ESLint started actually checking this file it really didn't like these property assignments to `window`. Resolves this.

.eslintrc.cjs
```diff
@@ -22,4 +22,16 @@ module.exports = {
     'object-shorthand': 'error',
     'prefer-arrow-callback': 'error',
   },
+  overrides: [
```
whiten, line 25: So I needed this to disable some of the warnings that Typescript flags now that `perf/**/*.ts and *.js` are included in the `tsconfig.json` `include` line. I had to include `*.js`, (or at least `runner.js`) in there to get around the below error. In any case, we can reduce this by fixing these errors, but I didn't want to bite that off now; I did fix some `eqeqeq` errors to avoid having to exclude that here too.

```
/Users/nate/dev/roci/replicache/perf/runner.js
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: perf/runner.js.
The file must be included in at least one of the projects provided
```

package.json
```diff
@@ -7,8 +7,8 @@
   "scripts": {
     "test": "web-test-runner",
     "test:watch": "web-test-runner --watch",
-    "format": "prettier --write '{doc,src,sample,perf}/**/*.{js,jsx,json,ts,tsx,html,css,md}' '*.{js,jsx,json,ts,tsx,html,css,md}'",
-    "check-format": "prettier --check '{doc,src,sample,perf}/**/*.{js,jsx,json,ts,tsx,html,css,md}' '*.{js,jsx,json,ts,tsx,html,css,md}'",
+    "format": "prettier --write '{doc,src,sample,perf}/**/*.{js,jsx,json,ts,tsx,html,css,md}' '*.{cjs,js,jsx,json,ts,tsx,html,css,md}'",
+    "check-format": "prettier --check '{doc,src,sample,perf}/**/*.{js,jsx,json,ts,tsx,html,css,md}' '*.{cjs,js,jsx,json,ts,tsx,html,css,md}'",
```
whiten, line 11: Added `.eslintrc.cjs` to prettier checks.

perf/data.js
```diff
@@ -16,16 +16,17 @@ export function randomData(type, len, datumSize) {
...
-    return randomString(len);
-  } else if (type == 'object') {
-    return randomObject(len);
-  } else if (type == 'arraybuffer') {
-    return randomUint8Array(len).buffer;
-  } else if (type == 'blob') {
-    return randomBlob(len);
-  } else {
-    throw new Error('unsupported');
+  switch (type) {
```
whiten, line 19: This responds to a review comment at https://github.com/rocicorp/replicache/commit/11426954d25c96077c49857c9abb7b3a5c07b6a7#r49230964, and I did it as all of these flagged `eqeqeq` previously.

rollup.config.js
```diff
@@ -6,7 +6,7 @@ import dts from 'rollup-plugin-dts';
 // We use esbuild for building the actual code.
 
 export default {
-  input: 'out/.dts/mod.d.ts',
+  input: 'out/.dts/src/mod.d.ts',
```
whiten, line 9: When I added files in `perf/` to the `tsconfig.json` include, this intermediate output file moved.

tsconfig.json
```diff
@@ -21,6 +21,6 @@
     "lib": ["dom", "esnext"],
     "preserveSymlinks": true
   },
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*.ts", "perf/**/*.ts", "perf/**/*.js"],
```
whiten, line 24: I believe the lack of `perf/**/*.ts` here before is why ESLint complained about `enum`, and also didn't catch all of the missing annotation errors that you hinted at.

whiten: FWIW, I checked `out/replicache.d.ts` after this change, and couldn't detect any hints of these new files in there, so I believe this is okay, but this is definitely beyond my ken.

whiten:
@arv, I flagged you for a re-review just because the follow-ups I needed went pretty far beyond your comments, and I touched `tsconfig.json` stuff I don't understand now. Thanks!

perf/lock.ts
arv: From the type system's view there is no difference between the following two:

```ts
const f1 = async () => 42;
const f2 = () => Promise.resolve(42);
```

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgRhgXhgQwgTzMGAKAShQD4YAWAJgG4AoUSWOClfI5UgBQCcQBbASwgBTAHRchEEABsAbkLyUCtOuGjwAzAC4YUTAAchIOPCSpEtemrhltug0ZMsmF1YwCst-YeNMnCFwzwAGye9j7MZtRAA

arv: Something else must be the issue.

.eslintrc.cjs
```diff
@@ -22,4 +22,16 @@ module.exports = {
...
+    {
+      files: ['perf/*.js'],
+      env: {
+        browser: true,
+      },
+      rules: {
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
+        '@typescript-eslint/ban-ts-comment': 'off',
```
arv: Maybe I'm misreading this?

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-ts-comment.md

But it seems like the default are all reasonable.

.eslintrc.cjs
arv: I would also be fine adding `.js` files to the main section. It just wasn't needed before.

.eslintrc.cjs
```diff
@@ -22,4 +22,16 @@ module.exports = {
...
     'prefer-arrow-callback': 'error',
   },
+  overrides: [
+    {
+      files: ['perf/*.js'],
+      env: {
+        browser: true,
+      },
+      rules: {
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
```
arv, line 32: I liked this rule. How was it getting in your way?

FWIW, You can always disable a rule on a specific line with a comment.

perf/lock.ts
```diff
@@ -18,34 +18,45 @@ export function benchmarks() {
...
 
-export async function isHeld(name: string) {
-  return await navigator.locks.request(
+export async function isHeld(name: string): Promise<boolean> {
+  return (await navigator.locks.request(
     name,
     {ifAvailable: true},
     async lock => lock === null,
-  );
+  )) as unknown as boolean;
```
arv, line 26: Not sure why this is needed. Looks like the d.ts was meant to deal with this.

🤷

perf/lock.ts
```diff
@@ -18,34 +18,45 @@ export function benchmarks() {
...
+    id: '',
     setup() {
       this.id = uuid();
     },
     async run() {
-      await navigator.locks.request(this.id, {mode: 'shared'}, () => undefined);
+      await navigator.locks.request(
+        this.id,
+        {mode: 'shared'},
+        async () => undefined,
```
arv, line 58: This one makes sense since it is not returning a Promise.

perf/runner.js
```diff
@@ -87,7 +87,7 @@ async function main() {
   await fs.rm(userDataDir, {recursive: true});
 
   // context.close does not terminate! Give it a second.
-  Promise.race([context.close(), wait(1000)]);
+  await Promise.race([context.close(), wait(1000)]);
```
arv, line 89: oops

rollup.config.js
arv: OK.

tsconfig.json
arv: sgtm

arv (approved):

.eslintrc.cjs
whiten: So prior to this PR, files in `perf/` got their ESLint config from `perf/.eslintrc.json`, and *didn't* inherit from the stuff in this file because `root: true` was specified there. Once I remove that and added those files to `tsconfig.json`, all of the "default" rules above *do* apply to `.js` files.

This override section is then relaxing a few of those rules, because the existing `perf/*.js` code violates them wildly. I think we can remove these overridden rules as we migrate JS files in perf to TS and meet the expectations we've set above, but it's ~38 errors right now, and I didn't want to fix them all just to introduce Typescript into the `perf/` directory.

This override also relaxes these rules *just* for JS files in `perf/`, not for any other files, so we keep the same tightness elsewhere we had previously.

.eslintrc.cjs
whiten: There were 9 `@ts-ignore` comments in `perf/*.js` files that are now being flagged. The default from `ban-ts-comment` is that you may never use `@ts-ignore`, but you *can* use `@ts-expect-err` if you put a non-trivial comment after it. It turns out I could just remove all of these `@ts-ignore` lines and this exclusion, which I've now done.

.eslintrc.cjs
whiten: It just fires for all of the unannotated JS code in `perf/*.js`, now that that falls under the rules in `.eslintrc.cjs`; this exclusion is only active for those files.

perf/lock.ts
whiten: The annotation is slightly off--it says that `request()` returns `undefined`, but really it returns the result of the inner callback.

PR #523, opened 2021-09-16
Switch to setup-node@v2 for built-in npm caching

whiten:


arv:
It is a lot cleaner so LGTM

It does not seem to impact the time much thogh

arv:
Are you still working on this one?

.github/workflows/js.yml
```diff
@@ -35,12 +39,14 @@ jobs:
...
+      PLAYWRIGHT_BROWSERS_PATH: 0
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: 14.x
           cache: 'npm'
       - run: npm install
-      - run: npx playwright install-deps
+      - run: npx playwright install-deps chromium
```
arv: That does not seem right. We run the tests on Firefox and WebKit too

arv (changes requested):

.github/workflows/js.yml
whiten: Yeah, I mixed this up with perf tests, where we default to only chromium.

PR #527, opened 2021-09-17
Implement proper CLI arg parsing for perf.

whiten:
Use this to support various browsers and output styles. All defaults
should be exactly as previously.

Example commands:

% npm run perf
% npm run perf -- --help
% npm run perf -- --devtools
% npm run perf -- --groups lock
% npm run perf -- --browsers firefox webkit
% npm run perf -- --browsers all
% npm run perf -- --format replicache


perf/lock.ts
```diff
@@ -67,6 +68,9 @@ function benchmarkHeldRead(mode: Mode) {
     group: 'lock',
     id: '',
     tab: null as Tab | null,
+    async skip() {
```
arv: doesn't look like you need `async` here

perf/runner.js
```diff
@@ -8,13 +10,101 @@ import * as os from 'os';
...
+    err.value = arg;
+    err.optionName = '--browsers';
+    throw err;
+  }
+  return arg;
+}
+
+function browserName(browser) {
+  let name = browser[0].toUpperCase() + browser.substr(1);
+  if (name === 'webkit') {
```
arv: never true

src/test-util.ts
```diff
@@ -59,6 +59,13 @@ export function defineTestFunctions(self: Window & typeof globalThis): void {
   }
 }
 
+export function isFirefox(): boolean {
+  return true;
```
arv: please fix

arv (approved):
with nits

src/test-util.ts
whiten: Yikes, thank you.

perf/runner.js
whiten: Fixed.

perf/lock.ts
whiten: Removed.

whiten: Removed, thank you.

PR #562, opened 2021-09-26
Update base next.js app location

joe-trellick:
Looks like they refactored their sample repos, this appears to work in my quick test


aboodman:
Thank you! This must have literally just changed -- I run these instructions regularly. I think the last time I ran it was not more than 2 weeks ago.

Issue #570, opened 2021-09-27
docs: integration guide: document why we `await` `sendPoke()`


aboodman:
... it's because without it, next sometimes terminates the call:

https://discord.com/channels/830183651022471199/830183651022471202/879416853087281252

Same here:

https://discord.com/channels/830183651022471199/830183651022471202/879416570231803924

Also update https://github.com/rocicorp/replicache-sample-chat

PR #579, opened 2021-09-27
Add comment explaining why we `await` on `sendPokes()`.

aboodman:
Toward https://github.com/rocicorp/replicache/issues/570


arv:
LGTM

PR #599, opened 2021-09-30
fixed repo url

ericmand:
The repo name and URL appears to have changed


ericmand:
I should add that I wasn't able to complete the subsequent steps of the getting started experience with the new URL, so while I think it is right, I'm not sure.

ericmand:
I just signed the contributor agreement but you can also submit yourself. Whatever you like.

aboodman:
Thanks, this is live now!

PR #603, opened 2021-09-30
Add a couple perf tests for small localStorage I/O.

whiten:
I swapped the order of firefox and webkit in the allBrowsers list
because runner.js would fail to start Firefox in the third position when
running with --browsers all, but it runs fine in the second. ¯\_(ツ)_/¯

### DIFF
--- a/perf/storage.ts
+++ b/perf/storage.ts
@@ -1,0 +1,46 @@
+import {uuid} from '../src/sync/uuid';
+import {randomString} from './data';
+import type {Benchmark} from './perf';
+
+export function benchmarks(): Benchmark[] {
+  return [localStorageRead(), localStorageWrite()];
+}
+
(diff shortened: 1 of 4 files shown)


perf/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+export function benchmarks(): Benchmark[] {
+  return [localStorageRead(), localStorageWrite()];
+}
+
+function localStorageRead() {
+  return {
+    name: 'localStorage read',
+    group: 'storage',
+    key: '',
+    value: <string | null>null,
```
arv, line 14:
```suggestion
    value: null as (string | nul),
```

arv: You can also use closure variables if you want.

arv (approved):

PR #612, opened 2021-10-21
refactor: Update Replicache class to no longer implement ReadTransaction, deprecate methods

### LINKED ISSUE #541: RFE: Expose clientID on Read/WriteTransaction

grgbkr:
Remove impl of `ReadTransaction` interface from `Replicache` class because while `Replicache` exposing these methods is slightly convenient

1. `Replicache`'s implementation is not actually transactional and thus cannot safely be passed to any function expecting a transaction
2. it makes it difficult to add a non-`Promise` property for `clientID` to the `ReadTransaction` interface (see #541)

BREAKING CHANGE: Replicache class no longer support RaedTransaction interface.  Users will
need to update code to use ReadTransaction via Replicache#query.

Refactor needed to address issue #541

### DIFF
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -120,9 +120,7 @@ type UnknownSubscription = Subscription<JSONValue | undefined, unknown>;
 type SubscriptionSet = Set<UnknownSubscription>;
 
 // eslint-disable-next-line @typescript-eslint/ban-types
-export class Replicache<MD extends MutatorDefs = {}>
-  implements ReadTransaction
-{
+export class Replicache<MD extends MutatorDefs = {}> {
   /** The URL to use when doing a pull request. */
   pullURL: string;
 
@@ -502,17 +500,26 @@ export class Replicache<MD extends MutatorDefs = {}>
     }
   }
 
-  /** Get a single value from the database. */
+  /**
+   * Get a single value from the database.
+   * @deprecated Use [[query]] instead.
+   */
   get(key: string): Promise<ReadonlyJSONValue | undefined> {
     return this.query(tx => tx.get(key));
   }
 
-  /** Determines if a single `key` is present in the database. */
+  /**
+   * Determines if a single `key` is present in the database.
+   * @deprecated Use [[query]] instead.
+   */
   has(key: string): Promise<boolean> {
     return this.query(tx => tx.has(key));
   }
 
-  /** Whether the database is empty. */
+  /**
+   * Whether the database is empty.
+   * @deprecated Use [[query]] instead.
+   */
   isEmpty(): Promise<boolean> {
     return this.query(tx => tx.isEmpty());
   }
@@ -525,6 +532,7 @@ export class Replicache<MD extends MutatorDefs = {}>
    * If `options` has an `indexName`, then this does a scan over an index with
    * that name. A scan over an index uses a tuple for the key consisting of
    * `[secondary: string, primary: string]`.
+   * @deprecated Use [[query]] instead.
    */
   scan<Options extends ScanOptions, Key extends KeyTypeForScanOptions<Options>>(
     options?: Options,
--- a/src/worker-test.ts
+++ b/src/worker-test.ts
@@ -71,6 +71,5 @@ async function testGethasScanOnEmptyDB(name: string, useMemstore = false) {
     expect(scanItems).to.have.length(0);
   }
 
-  await t(rep);
   await rep.query(t);
 }


src/replicache.ts
```diff
@@ -502,45 +498,6 @@ export class Replicache<MD extends MutatorDefs = {}>
     }
   }
 
-  /** Get a single value from the database. */
```
aboodman, line 505: Could we leave the methods, but just not declare that we implement the interface?

aboodman: (I don't know that anyone is really using this API, but it seems like a bigger question to remove it).

grgbkr: Sure.  We could leave all of them if they are useful, but it may also be somewhat confusing to have nearly all the methods of ReadTransaction but not actually implement ReadTransaction (and we will likely forget to add any additions to ReadTransaction here).

arv: I prefer removing them but I'd be OK with marking them as deprecated for one major release cycle.

src/replicache-types.test.ts
```diff
@@ -129,43 +129,6 @@ test.skip('Test register param [type checking only]', async () => {
   });
 });
 
-// Only used for type checking
-test.skip('Key type for scans [type checking only]', async () => {
```
arv: These type tests are not covered on ReadTransaction as far as I can tell. Can you refactor to wrap it in a `query`.

src/replicache-types.test.ts
```diff
@@ -306,26 +269,3 @@ test.skip('mut [type checking only]', async () => {
     rep.mutate.abc(1, 2, 3);
   }
 });
-
-// Only used for type checking
-test.skip('scan with index [type checking only]', async () => {
```
arv: Same here?

src/replicache.ts
aboodman: That seems like a good solution.

src/replicache.ts
```diff
@@ -502,45 +498,6 @@ export class Replicache<MD extends MutatorDefs = {}>
...
-    options?: Options,
-  ): ScanResult<Key, ReadonlyJSONValue> {
-    return new ScanResult<Key>(
-      options,
-      async () => {
-        await this._ready;
-        const dagRead = await this._dagStore.read();
-        return db.readFromDefaultHead(dagRead);
-      },
-      true, // shouldCloseTransaction
```
arv: Followup: Remove `shouldCloseTransaction`

src/replicache.test.ts
```diff
@@ -1537,9 +1537,8 @@ testWithBothStores('pullInterval in constructor', async () => {
 });
 
 testWithBothStores('closeTransaction after rep.scan', async () => {
```
arv: I'm not sure this test makes any sense any more.

Yeah. Let's remove this test.

arv (changes requested):

aboodman:
PS: Just to flesh out the discussion:

> but it may also be somewhat confusing to have nearly all the methods of ReadTransaction but not actually implement ReadTransaction (and we will likely forget to add any additions to ReadTransaction here).

I'm not that concerned about it being confusing. I don't think people would look at docs of ReadTransaction then compare to docs of Replicache and wonder what's going on. And the type system won't let you somehow make a mistake.

But I do agree that these methods aren't that useful and just extra stuff for us to maintain, and probably also foot guns (non transactional) so let's remove.

aboodman:
> so let's remove.

err, deprecate, then remove.

src/replicache.ts
grgbkr: cool, i thought about the deprecation approach, but wasnt sure how we were handling api breakages with releases at this stage.  ill mark deprecated and file a follow up to remove.  how often is a major release?

aboodman:
We don't have a fixed schedule (I would like to though).

In practice it's been ~monthly.

grgbkr:
Update `Replicache` class to no longer implement `ReadTransaction` because while `Replicache` exposing these methods is slightly convenient

1. `Replicache`'s implementation is not actually transactional and thus cannot safely be passed to any function expecting a transaction
2. it makes it difficult to add a non-`Promise` property for `clientID` to the `ReadTransaction` interface (see #541).

Currently does not remove `ReadTransaction` methods from `Replicache`, but marks them as
deprecated, pointing to `Replicache#query`.  These methods will be removed in the
next major release.

No longer a breaking change.

aboodman (approved):

src/replicache-types.test.ts
grgbkr: Will do in the follow up when i actually remove the methods.

src/replicache-types.test.ts
grgbkr: Will do in the follow up when i actually remove the methods.

src/replicache.ts
grgbkr: Will do in the follow up when i actually remove the methods.

src/replicache.test.ts
grgbkr: agree Will do in the follow up when i actually remove the methods.

src/replicache.ts
```diff
@@ -502,17 +500,26 @@ export class Replicache<MD extends MutatorDefs = {}>
     }
   }
 
-  /** Get a single value from the database. */
+  /**
+   * Get a single value from the database.
+   * @deprecated Use [[query]] instead.
```
arv, line 505: I feel like we should comment when we deprecated them. It makes it easier to reason about when it is safe to remove them later.

This can be a non jsdoc comment. Something like:
```
// Deprecated after 7.0.0
```

arv (commented):
LGTM

PR #613, opened 2021-10-21
feat: Expose clientID on ReadTransaction

### LINKED ISSUE #541: RFE: Expose clientID on Read/WriteTransaction

grgbkr:
Exposes clientID on ReadTransaction (and subclasses, e.g. WriteTransaction, IndexTransaction, etc).

Closes #541

### DIFF
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -137,7 +137,7 @@ testWithBothStores('get, has, scan on empty db', async () => {
     expect(scanItems).to.have.length(0);
   }
 
-  await t(rep);
+  await rep.query(t);
 });
 
 testWithBothStores('put, get, has, del inside tx', async () => {
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -2805,3 +2805,26 @@ test('mutate args in mutation', async () => {
     v: 1,
   });
 });
+
+test('client ID is set correctly on transactions', async () => {
+  const store = new TestMemStore();
+  const rep = await replicacheForTesting(
+    'client-id-is-set-correctly-on-transactions',
+    {
+      experimentalKVStore: store,
+      mutators: {
+        async expectClientID(tx, args: {expectedClientID: string}) {
+          expect(tx.clientID).to.equal(args.expectedClientID);
+        },
+      },
+    },
+  );
+
+  const repClientID = await rep.clientID;
+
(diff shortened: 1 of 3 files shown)


src/replicache.test.ts
```diff
@@ -2805,3 +2805,26 @@ test('mutate args in mutation', async () => {
     v: 1,
   });
 });
+
+test('client ID is set correctly on transactions', async () => {
+  const store = new TestMemStore();
+  const rep = await replicacheForTesting(
+    'client-id-is-set-correctly-on-transactions',
+    {
+      experimentalKVStore: store,
```
arv, line 2814: skip this and use `testWithBothStores`

`TestMemStore` has some logging stuff in it that a very few select tests uses.

src/replicache.test.ts
```diff
@@ -2805,3 +2805,26 @@ test('mutate args in mutation', async () => {
...
+      mutators: {
+        async expectClientID(tx, args: {expectedClientID: string}) {
```
arv, line 2816: or just:

```ts
async expectClientID(tx, expectedClientID: string) {
```

src/replicache.ts
```diff
@@ -1078,6 +1080,7 @@ export class Replicache<MD extends MutatorDefs = {}> {
         originalHash = rebaseOpts.original;
       }
 
+      const clientID = await this._clientIDPromise;
```
arv, line 1083: move down closer to usage

arv (commented):
LGTM

src/replicache.ts
grgbkr: Both clientID and dbWrite are passed as args to WriteTransactionImpl.  I just await them here in the order they are passed to WriteTransactionImpl, and nothing else is between the usage.

I could inline this
const tx = new WriteTransactionImpl(await this._clientIDPromise, dbWrite, this._lc);

but I don't think it helps readability much.

src/replicache.test.ts
grgbkr: got it

src/replicache.ts
arv: I just felt that it should not be done before `newLocal` since it is not needed there.

`_clientIDPromise` is guaranteed to have been resolved here so it is not a performance issue.

PR #617, opened 2021-10-22
fix: Fix deadlock/transaction issues with expose clientID on Read/WriteTransaction

### LINKED ISSUE #541: RFE: Expose clientID on Read/WriteTransaction

grgbkr:
This change

1. Addresses arv's feedback on PR #613
2. Fixes issue where awaiting clientID inside a dag transaction was causing deadlocks and/or prematurely closing idb transactions.
3. Restructures `Replicache#_open` to avoid unnecessarily delaying the resolution of _clientIDPromise (currently blocks on db initialization and `this._getRoot`).

Closes #541

### DIFF
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -142,7 +142,6 @@ export class Replicache<MD extends MutatorDefs = {}> {
   private _online = true;
   private readonly _logger: Logger;
   private readonly _ready: Promise<void>;
-  private readonly _resolveReady: () => void;
   private readonly _clientIDPromise: Promise<string>;
   private _root: Promise<string | undefined> = Promise.resolve(undefined);
   private readonly _mutatorRegistry = new Map<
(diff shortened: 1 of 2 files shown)


src/replicache.ts
```diff
@@ -333,9 +337,10 @@ export class Replicache<MD extends MutatorDefs = {}> {
       sync.initClientID(this._kvStore),
       db.maybeInitDefaultDB(this._dagStore),
     ]);
+    resolveClientID(clientID);
 
     // Now we have both a clientID and DB!
-    this._resolveReady();
+    resolveReady();
```
arv, line 344: We might be able to only use one resolver now

src/replicache.ts
```diff
@@ -573,13 +577,14 @@ export class Replicache<MD extends MutatorDefs = {}> {
     f: (tx: IndexTransactionImpl) => Promise<void>,
   ): Promise<void> {
     await this._ready;
+    // clientID must be awaited ouside dag transaction to avoid a premature
```
grgbkr, line 581: @arv do you think these comments are useful?

arv:
LGTM

src/replicache.ts
arv: yes

arv: but if we make _ready and clientID promise the same promise this would be clearer maybe?

src/replicache.ts
grgbkr: the ready bit is still used in a bunch of places where clientID is not at the moment, so its not straightforward to get rid of ready.

grgbkr: i also made another small change that avoids blocking resolution of the clientIdPromise on the db initialization

aboodman:
Nice!

aboodman:
I can't see exactly what went wrong here with the perf test, but I can replicate it locally so I'm going to revert, sorry!

PR #620, opened 2021-10-24
New benchmarkWriteSubRead

aboodman:


perf/replicache.ts
```diff
@@ -155,193 +152,107 @@ export function benchmarkCreateIndex(opts: {
...
-  const maxCount = 1000;
-  const minCount = 10;
+  const {valueSize, numSubsTotal, keysPerSub, keysWatchedPerSub, numSubsDirty} =
+    opts;
+
+  const numKeys = keysPerSub * numSubsTotal;
+  const cacheSizeMB = (numKeys * valueSize) / 1024 / 1024;
+  const kbReadPerSub = (keysWatchedPerSub * valueSize) / 1024;
+
+  const data = Array.from({length: numKeys}).map(() => randomString(valueSize));
```
arv, line 177: Can we change this to randomJSON?

arv (commented):
LGTM

Thanks

perf/replicache.ts
aboodman: I did that in a subsequent commit. It didn't make a huge difference with a realistic number of keys.

arv: I have some ideas for improving hashing so I think that it would be useful to have more realistic data.

PR #621, opened 2021-10-24
Create a randomObject instead of randomString.

aboodman:
The test is very sensitive to number of keys in the object. I chose 16 as that seemed like a reasonable 90th pct of fields.

### DIFF
--- a/perf/data.ts
+++ b/perf/data.ts
@@ -25,10 +25,10 @@ function randomDatum(type: RandomDataType, len: number): RandomDatum {
   }
 }
 
-function randomObject(len: number): Record<string, string> {
+export function randomObject(len: number): Record<string, string> {
   const ret: Record<string, string> = {};
-  for (let i = 0; i < Math.min(100, len); i++) {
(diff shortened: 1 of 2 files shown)


perf/data.ts
```diff
@@ -25,10 +25,10 @@ function randomDatum(type: RandomDataType, len: number): RandomDatum {
   }
 }
 
-function randomObject(len: number): Record<string, string> {
+export function randomObject(len: number): Record<string, string> {
   const ret: Record<string, string> = {};
-  for (let i = 0; i < Math.min(100, len); i++) {
-    ret[`k${i}`] = randomString(Math.ceil(len / 100));
+  for (let i = 0; i < Math.min(16, len); i++) {
+    ret[`k${i}`] = randomString(Math.ceil(len / 16));
```
arv, line 31: strings all the way down

arv: Maybe use a combination of numbers, arrays, booleans etc

PR #624, opened 2021-10-26
Update indexes incrementally after pull.

### LINKED ISSUE #601: Full index rebuild on first pull after index creation
The first time we pull after creating an index, that index will be fully rebuilt. The relevant code is here:

https://github.com/rocicorp/replicache/blob/main/src/sync/pull.ts#L162

The reason is somewhat subtle. Imagine the situation in which we have just created an index, then pull:

```
S1(lmid=0) - M1 - CI1 - main
    \ S2(lmid=1) - sync
```

If we simply apply the patch from `S2` to `S1` and commit, then the index created by `CI1` will be lost completely. It's not defined yet at `S1`.

However, we also cannot just take the index state from `CI1` and apply the patch from `S2` because the patch at `S2` is with respect to `S1`, not `CI1`. In the simple case where `S2` includes only the changes from `M1` then this might appear to work because applying the patch from `S2` to the index state at `CI1` will result in no visible change. However imagine the case where the authoritative db computed a different result for `M1` -- for example, maybe it ignored this change. In that case, the index state at `CI1` reflects `M1` but the patch from `S2` does not. If we reuse the index state from `CI1` then we will exit this pull with the index having entries in it that the main map does not.

To avoid this issue what we do now is simply rebuild all the indexes described by `M1` on top of `S1` before applying the patch from `S2`. Our index creation code ignores duplicate index definitions (https://github.com/rocicorp/replicache/blob/main/src/db/write.ts#L209) so in practice this means we will build indexes introduced between `S1` and `main`, but not other indexes.

Anyway, this is kind of a bummer. I believe that we can now fix this by:

1. start with the index state at `main`.
2. compute a diff from `main` to `snapshot`.
3. modify the index state to reflect each difference

Our diff computation is currently slow but this should still be faster than recreating the index from scratch, and once #596 is implemented it will get even faster.

aboodman:
Fixes #601

### DIFF
--- a/src/db/write.ts
+++ b/src/db/write.ts
@@ -14,6 +14,7 @@ import {IndexWrite, IndexOperation, indexValue, IndexRead} from './index';
 import type {LogContext} from '../logger';
 import {BTreeRead, BTreeWrite} from '../btree/mod';
 import {asyncIterableToArray} from '../async-iterable-to-array';
+import {lazy} from '../lazy';
 
 type IndexChangeMeta = {
   type: MetaType.IndexChange;
@@ -137,25 +138,8 @@ export class Write extends Read {
     if (this._meta.type === MetaType.IndexChange) {
       throw new Error('Not allowed');
     }
-    const oldVal = await this.map.get(key);
-    if (oldVal !== undefined) {
-      await updateIndexes(
-        lc,
-        this.indexes,
-        this._dagWrite,
-        IndexOperation.Remove,
-        key,
-        oldVal,
-      );
-    }
-    await updateIndexes(
-      lc,
-      this.indexes,
-      this._dagWrite,
-      IndexOperation.Add,
-      key,
-      val,
-    );
+    const oldVal = lazy(() => this.map.get(key));
+    await updateIndexes(lc, this.indexes, this._dagWrite, key, oldVal, val);
 
     await this.map.put(key, val);
   }
@@ -166,15 +150,15 @@ export class Write extends Read {
     }
 
     // TODO(arv): This does the binary search twice. We can do better.
-    const oldVal = await this.map.get(key);
+    const oldVal = lazy(() => this.map.get(key));
     if (oldVal !== undefined) {
       await updateIndexes(
         lc,
         this.indexes,
         this._dagWrite,
-        IndexOperation.Remove,
         key,
         oldVal,
+        undefined,
       );
     }
     return this.map.del(key);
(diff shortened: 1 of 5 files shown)


aboodman:
Right now this uses the slow linear diff, but can trivially be replaced by the fast diff when #610 lands.

aboodman:
Existing tests cover the semantics. I manually verified that we only do the incremental rebuild using the todo app.

src/lazy.ts
```diff
@@ -0,0 +1,9 @@
+export function lazy<T>(factory: () => T): () => T {
+  let value: T | undefined;
+  return () => {
+    if (value === undefined) {
```
arv, line 4: What if `T` includes `undefined`?

src/db/write.ts
```diff
@@ -166,15 +150,15 @@ export class Write extends Read {
     }
 
     // TODO(arv): This does the binary search twice. We can do better.
-    const oldVal = await this.map.get(key);
+    const oldVal = lazy(() => this.map.get(key));
     if (oldVal !== undefined) {
```
arv, line 154: Is it ever undefined now?

src/db/write.ts
```diff
@@ -137,25 +138,8 @@ export class Write extends Read {
...
-      this.indexes,
-      this._dagWrite,
-      IndexOperation.Add,
-      key,
-      val,
-    );
+    const oldVal = lazy(() => this.map.get(key));
+    await updateIndexes(lc, this.indexes, this._dagWrite, key, oldVal, val);
 
     await this.map.put(key, val);
```
arv, line 144: We can make put return the old value while we are in there

src/db/write.ts
```diff
@@ -166,15 +150,15 @@ export class Write extends Read {
...
         lc,
         this.indexes,
         this._dagWrite,
-        IndexOperation.Remove,
         key,
         oldVal,
+        undefined,
       );
     }
     return this.map.del(key);
```
arv, line 164: Can we change the order here? `del` returns if something was deleted or not so we should not need to do a get.

src/lazy.ts
arv: then we keep calling it... I guess it is fine but I think what you are aiming for is to memoize the result

src/sync/pull.ts
```diff
@@ -140,36 +141,57 @@ export async function beginPull(
...
+    const lastIntegratedMap = new BTreeRead(dagRead, lastIntegrated.valueHash);
+
+    for await (const change of dbWrite.map.diff(lastIntegratedMap)) {
+      await updateIndexes(
+        lc,
+        dbWrite.indexes,
+        dagWrite,
+        change.key,
+        () =>
+          Promise.resolve(
```
arv, line 188: Is it really worth this extra complication?

aboodman: Well the only reason we need to load the old val is to update indexes. But if we're making a change that doesn't affect any indexes then it seems sad to load it for no reason? Unless we think it's likely to already be cached which I guess maybe it is.

arv:
This needs more work. Just writing that down so that it does not get lost

PR #626, opened 2021-10-26
fix: Do not send push or pull requests when pushURL/pullURL are not provided

### LINKED ISSUE #625: Bug: Push and pull request times are being included in perf tests
These tests run without a server implementing the push and pull endpoints, and so these requests fail with a 404.  These tests do not intend to measure the performance of push and pull, however the latency of these failed request is being included in some of the tests' timings.

image

grgbkr:
Updates `Replicache`s logic to not send push requests when `ReplicacheOptions.pushURL` is not provided (unless `ReplicacheOptions.pusher` is provided), and similarly to not send pull requests when `ReplicacheOptions.pullURL` is not provided (unless `ReplicacheOptions.puller` is provided).

Implementation uses the null object pattern , using a `NullConnectionLoop` when the corresponding url is not specified.

This fixes an issue where push and pull requests were being sent in the performance tests, unintentionally impacting the timing results. 

This made a very large difference in the persistent store tests for read and scan. 

Prior to this change
```
read tx 1024x1000 x 9.74 MB/s ±78.9% (7 runs sampled)
read tx 1024x5000 x 9.05 MB/s ±83.0% (5 runs sampled)
scan 1024x1000 x 9.84 MB/s ±80.2% (8 runs sampled)
scan 1024x5000 x 57.11 MB/s ±593.0% (5 runs sampled)
```

After this change
```
read tx 1024x1000 x 70.77 MB/s ±37.0% (35 runs sampled)
read tx 1024x5000 x 85.97 MB/s ±36.4% (9 runs sampled)
scan 1024x1000 x 72.34 MB/s ±66.7% (36 runs sampled)
scan 1024x5000 x 95.74 MB/s ±52.9% (10 runs sampled)
```

Closes #625

### DIFF
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -68,6 +68,7 @@ async function replicacheForTesting<MD extends MutatorDefs = {}>(
   });
 }
 
+// eslint-disable-next-line @typescript-eslint/ban-types
 async function replicacheForTestingNoDefaultURLs<MD extends MutatorDefs = {}>(
   name: string,
   {
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -57,10 +57,23 @@ let overrideUseMemstore = false;
 // eslint-disable-next-line @typescript-eslint/ban-types
 async function replicacheForTesting<MD extends MutatorDefs = {}>(
   name: string,
+  options: ReplicacheOptions<MD> = {},
+): Promise<ReplicacheTest<MD>> {
+  const pullURL = 'https://pull.com/?name=' + name;
+  const pushURL = 'https://push.com/?name=' + name;
+  return replicacheForTestingNoDefaultURLs(name, {
+    pullURL,
+    pushURL,
+    ...options,
+  });
+}
+
+async function replicacheForTestingNoDefaultURLs<MD extends MutatorDefs = {}>(
+  name: string,
   {
-    pullURL = 'https://pull.com/?name=' + name,
+    pullURL,
     pushDelay = 60_000, // Large to prevent interfering
-    pushURL = 'https://push.com/?name=' + name,
+    pushURL,
     useMemstore = overrideUseMemstore,
 
     ...rest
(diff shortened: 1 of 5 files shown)


src/replicache.ts
```diff
@@ -297,21 +303,25 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
-    );
-
-    this._pushConnectionLoop = new ConnectionLoop(
-      new PushDelegate(
-        this,
-        () => this._invokePush(MAX_REAUTH_TRIES),
-        getLogger(['PUSH'], logLevel),
-      ),
-    );
+    this._pullConnectionLoop = this.pullURL
```
aboodman: Does this still work if the user has a custom `puller` or `pusher` specified: https://doc.replicache.dev/api/classes/Replicache#puller?

I think we only want to use the null loop in the case where both the puller is null and the pullURL is empty.

aboodman: Also I guess we need a unit test that exercises a custom puller/pusher 😳?

grgbkr: good catch, had not looked at puller/pusher.  updating to handle.

src/replicache.ts
```diff
@@ -297,21 +303,27 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+            new PullDelegate(
+              this,
+              () => this._invokePull(),
+              getLogger(['PULL'], logLevel),
+            ),
+          )
+        : createNullConnectionLoop();
+
+    this._pushConnectionLoop =
+      options.pushURL || options.pusher
```
aboodman, line 318: This kind of combination always weirds me out because I can not remember the precedence off hand. Does eslint whine if you put parens around `options.pushURL || options.pusher` ? If it's ok with eslint, can we add. If not, shrug.

aboodman (approved):

src/replicache.ts
grgbkr: prettier strips them out

src/replicache.test.ts
```diff
@@ -80,6 +94,7 @@ async function replicacheForTesting<MD extends MutatorDefs = {}>(
   await rep.clientID;
   fetchMock.post(pullURL, {lastMutationID: 0, patch: []});
   fetchMock.post(pushURL, 'ok');
+  console.log(name + ' ' + pushURL);
```
arv, line 97: remove

arv:
LGTM

src/replicache.test.ts
grgbkr: thanks for catching, removing in https://github.com/rocicorp/replicache/pull/631

PR #637, opened 2021-10-28
Bump

aboodman:


aboodman:
Merge `bump` from `repc` to Replicache to make releases a bit easier. This imports the entire history of repc so we don't lose our precious history of the bump tool :-).

aboodman:
> This imports the entire history of repc so we don't lose our precious history of the bump tool :-).

Though note that the history was probably already in the repo from some other merge, since I have a bias to never throw anything away.

PR #645, opened 2021-10-29
fix: update perf tests with ops/sec units to output median rather than mean

grgbkr:
This makes them more consistent with tests with MBs/sec output (which already use median), and should also make them more stable as results can't be greatly skewed by a small number of outliers.

### DIFF
--- a/perf/perf.ts
+++ b/perf/perf.ts
@@ -94,7 +94,8 @@ async function runBenchmark(
     return formatAsBenchmarkJS({
       name: benchmark.name,
       value:
-        medianBytesPerSecond || `${((1 / medianTime) * 1000).toFixed(2)} ops/sec `,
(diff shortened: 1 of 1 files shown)


aboodman (approved):
thanks!

perf/perf.ts
```diff
@@ -93,7 +93,8 @@ async function runBenchmark(
       medianTime;
     return formatAsBenchmarkJS({
       name: benchmark.name,
-      value: bytesPerSecond || `${((runs / sum) * 1000).toFixed(2)} ops/sec `,
+      value:
+        bytesPerSecond || `${((1.0 / medianTime) * 1000).toFixed(2)} ops/sec `,
```
arv: `1.0` -> `1`

arv (approved):
Thanks

PR #646, opened 2021-10-29
feat: Improve replicache output format for perf tests.

grgbkr:
Adds:
1. avg ms/op (in addiiton to existing percentiles)
2. percentiles for MB/s (before just median)
3. avg for MB/s
4. clearer labeling/spacing

Sample output line
```
populate 1024x1000 (clean, indexes: 0) 50/75/90/95%=146.80,151.20,178.50,178.50 ms/op avg=153.42 ms/op 50/75/90/95%=6.65,6.46,5.47,5.47 MB/s avg=6.37 MB/s (5 runs sampled)
```

### DIFF
--- a/perf/perf.ts
+++ b/perf/perf.ts
@@ -76,28 +76,49 @@ async function runBenchmark(
 
   const median = 0.5;
   const medianTime = times[Math.floor(runs * median)];
-  const bytesPerSecond = benchmark.byteSize
+  const medianBytesPerSecond = benchmark.byteSize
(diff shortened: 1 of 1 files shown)


aboodman:
Nice but is there a reason to include the ops/sec? Is anyone (@arv?) advocating for it?

arv (approved):
I agree with @aboodman -- I don't think we need ops/s in the replicache format. It just makes the output very long.

grgbkr:
I've rolled the parts worth keeping of this into https://github.com/rocicorp/replicache/pull/654, so going to abandon this pull request.

PR #662, opened 2021-11-04
Perf dashboard spruce

aboodman:
Only display the benchmarks from latest results.

### DIFF
--- a/perf-v2/index.html
+++ b/perf-v2/index.html
@@ -137,14 +137,15 @@
         function init() {
           function collectBenchesPerTestCase(entries) {
             const map = new Map();
+            for (const bench of entries[0].benches) {
+              map.set(bench.name, []);
+            }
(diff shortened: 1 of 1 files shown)


perf-v2/index.html
```diff
@@ -101,38 +106,46 @@
...
 
         function init() {
           function collectBenchesPerTestCase(entries) {
             const map = new Map();
+            for (const bench of entries[0].benches) {
+              map.set(bench.name, []);
+            }
             for (const entry of entries) {
-              const {commit, date, tool, benches} = entry;
+              const { commit, date, tool, benches } = entry;
```
arv, line 144: What prettier config are you using for all these changes? It is not consistent with the rest of the code base?

arv: and it is causing diff churn...

PR #663, opened 2021-11-04
Spruce up the benchmarks we run by default.

aboodman:
This trims it down to just the MemStore versions of:
- the write/sub/read family
- the populate family
- scan
- createIndex

We remove all the IDB tests (this is not how we plan to use IDB,
so it's irrelevant and confusing), and the read tests (does not represent
real-world usage).

### DIFF
--- a/perf/replicache.ts
+++ b/perf/replicache.ts
@@ -302,16 +302,6 @@ async function makeRepWithPopulate(useMemstore: boolean) {
 
 export function benchmarks(): Benchmark[] {
   const bs = (useMemstore: boolean) => [
-    benchmarkPopulate({numKeys: 1000, clean: true, useMemstore}),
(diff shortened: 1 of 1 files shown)


aboodman:
It also reorders the tests so they are in sort of order of priority.

perf/replicache.ts
```diff
@@ -340,6 +330,11 @@ export function benchmarks(): Benchmark[] {
       useMemstore,
     }),
     // 128 mb is unusable
```
phritz, line 332: whats the relevance of this comment at this point? none?

phritz (approved):

perf/replicache.ts
aboodman: 128 is our goal so I wanted to specifically note why it's not included. I'll make that more clear.

PR #665, opened 2021-11-04
feat: add p95 graphs to perf-v2 benchmarks

grgbkr:


arv (approved):
Nice. A lot simpler than I anticipated

PR #669, opened 2021-11-04
perf: Optimize subscriptions for scans with a limit

### LINKED ISSUE #666: Optimize subscriptions for scans with a limit
## Background 

Right now we do not use `limit` when determining if a `scan` is affected by a key. See [subscriptions.ts](https://github.com/rocicorp/replicache/blob/2d2340227184c058bcb49ed1070bf129eb7a0385/src/subscriptions.ts#L39). 

We only use the `startKey` and `prefix`. This means that if we have something like:

```js
rep.subscripe(tx => {
  for await (const entry of tx.scan({startKey: 'a', limit: 5}) {
     console.log(e);
  }
}, {onData() {});
```

we cannot tell whether a change to key `'x'` should affect the subscription function.

## Proposed solution

Inside `SubscriptionTransactionWrapper`'s `scan` method.
- If the `ScanOptions` have a `limit` and we yielded `limit` values then,
  - we keep track of what key that was associated with. Call this _endInclusiveKey_.
- We [currently only store](https://github.com/rocicorp/replicache/blob/2d2340227184c058bcb49ed1070bf129eb7a0385/src/transactions.ts#L124) the `ScanOptions` but we now also store an optional `endInclusiveKey` property and set it to _endInclusiveKey_ if we have it.
- Then in `scanOptionsMatchesKey` we also take into account `limit` and `endInclusiveKey`

grgbkr:
### Problem 

Right now we do not use `limit` when determining if a `scan` is affected by a key.(https://github.com/rocicorp/replicache/blob/2d2340227184c058bcb49ed1070bf129eb7a0385/src/subscriptions.ts#L39). 

We only use the `startKey` and `prefix`. This means that if we have something like:

```js
rep.subscripe(tx => {
  for await (const entry of tx.scan({startKey: 'a', limit: 5}) {
     console.log(e);
  }
}, {onData() {});
```

we cannot tell whether a change to key `'x'` should affect the subscription function.

### Solution

Inside `SubscriptionTransactionWrapper`'s `scan` method.
- We currently only store a `ScanOptions` for each scan, update to store a `ScanSubrscriptionInfo`, consisting of `ScanOptions` and the new field `inclusiveLimitKey`. `inclusiveLimitKey` will be populated based on a callback driven by the scan implementation in `btree/node.ts` (this callback has to be threaded through a few layers)
- Then in `scanOptionsMatchesKey` we know that a key does not match if there is a limit and the changed key is greater than the inclusiveLimitKey.  This works for both `prefix`, `startKey`, and a combination of `prefix` and `startKey`.  

Note this optimization is only applied when a subscription reads its scan to its limit (it choses to stop early, or it runs out of entries).   We can optimize some cases where a subscription does not read its scan to its limit, however these optimization are only correct if the subscription body is a pure function on the replicache store state.  Subscription bodies should be pure in this way, but so far replicache behavior is correct even if they are not pure. We chose to not implement these further optimizations, erroring on the side of correctness over performance.  We can of course revisit this if it turns out not reading a scan to its limit is a common use case.  

### Perf measurements

This greatly improves writeSubRead benchmark performance when random invalidates are used, **reducing the median time by ~70%**.  This benchmark uses 128 scans each using startKey and limit.   

Made on my M1 Max w 64 GB of memory

Before this optimization:
```
[MemStore] writeSubRead randomInvalidates false 100MB total, 128 subs total, 5 subs dirty, 16kb read per sub 50/75/90/95%=2.50/4.00/11.40/11.40 ms avg=4.56 ms (7 runs sampled)
[MemStore] writeSubRead randomInvalidates true 100MB total, 128 subs total, 5 subs dirty, 16kb read per sub 50/75/90/95%=9.50/12.70/19.70/19.70 ms avg=12.69 ms (7 runs sampled)
```

With this optimization:
```
Running 2 benchmarks on Chromium...
[MemStore] writeSubRead randomInvalidates false 100MB total, 128 subs total, 5 subs dirty, 16kb read per sub 50/75/90/95%=2.50/4.20/10.70/10.70 ms avg=4.49 ms (7 runs sampled)
[MemStore] writeSubRead randomInvalidates true 100MB total, 128 subs total, 5 subs dirty, 16kb read per sub 50/75/90/95%=3.00/4.40/11.10/11.10 ms avg=4.99 ms (7 runs sampled)
Done!
```

Closes #666

### DIFF
--- a/src/btree/node.ts
+++ b/src/btree/node.ts
@@ -221,7 +221,7 @@ export class DataNodeImpl extends NodeImpl<ReadonlyJSONValue> {
     prefix: string,
     fromKey: string,
     limit: number,
-    onLimitKey?: (key: string) => void,
+    onLimitKey?: (inclusiveLimitKey: string) => void,
   ): AsyncGenerator<Entry<ReadonlyJSONValue>, number, unknown> {
     const {entries} = this;
     let i = binarySearch(fromKey, entries);
@@ -439,7 +439,7 @@ export class InternalNodeImpl extends NodeImpl<Hash> {
     prefix: string,
     fromKey: string,
     limit: number,
-    onLimitKey?: (key: string) => void,
+    onLimitKey?: (inclusiveLimitKey: string) => void,
   ): AsyncGenerator<Entry<ReadonlyJSONValue>, number> {
     const {entries} = this;
     let i = binarySearch(fromKey, entries);
@@ -451,7 +451,7 @@ export class InternalNodeImpl extends NodeImpl<Hash> {
     }
     for (; i < entries.length && limit > 0; i++) {
       const childNode = await tree.getNode(entries[i][1]);
-      limit = yield* childNode.scan(tree, prefix, fromKey, limit, onKey);
+      limit = yield* childNode.scan(tree, prefix, fromKey, limit, onLimitKey);
     }
     return limit;
   }
--- a/src/btree/node.ts
+++ b/src/btree/node.ts
@@ -221,11 +221,7 @@ export class DataNodeImpl extends NodeImpl<ReadonlyJSONValue> {
     prefix: string,
     fromKey: string,
     limit: number,
-    onKey?: (
-      key: string,
-      isLastBeforeLimit: boolean,
-      isLastEntry: boolean,
-    ) => void,
+    onKey?: (key: string, isInclusiveLimit: boolean) => void,
   ): AsyncGenerator<Entry<ReadonlyJSONValue>, number, unknown> {
     const {entries} = this;
     let i = binarySearch(fromKey, entries);
@@ -238,7 +234,7 @@ export class DataNodeImpl extends NodeImpl<ReadonlyJSONValue> {
       limit--, i++
     ) {
       if (onKey) {
-        onKey(entries[i][0], limit === 1, i === entries.length - 1);
+        onKey(entries[i][0], limit === 1);
       }
       yield entries[i];
     }
@@ -443,11 +439,7 @@ export class InternalNodeImpl extends NodeImpl<Hash> {
     prefix: string,
     fromKey: string,
     limit: number,
-    onKey?: (
-      key: string,
-      isLastBeforeLimit: boolean,
-      isLastEntry: boolean,
-    ) => void,
+    onKey?: (key: string, isInclusiveLimit: boolean) => void,
   ): AsyncGenerator<Entry<ReadonlyJSONValue>, number> {
     const {entries} = this;
     let i = binarySearch(fromKey, entries);
--- a/src/btree/node.ts
+++ b/src/btree/node.ts
@@ -221,7 +221,7 @@ export class DataNodeImpl extends NodeImpl<ReadonlyJSONValue> {
     prefix: string,
     fromKey: string,
     limit: number,
-    onKey?: (key: string, isInclusiveLimit: boolean) => void,
+    onLimitKey?: (key: string) => void,
   ): AsyncGenerator<Entry<ReadonlyJSONValue>, number, unknown> {
     const {entries} = this;
     let i = binarySearch(fromKey, entries);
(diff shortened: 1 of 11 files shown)


src/transactions.ts
```diff
@@ -151,15 +158,21 @@ export class SubscriptionTransactionWrapper implements ReadTransaction {
...
   ): ScanResult<Key, ReadonlyJSONValue> {
-    this._scans.push(toDbScanOptions(options));
-    return this._tx.scan(options);
+    const scanInfo: ScanSubrscriptionInfo = {
+      options: toDbScanOptions(options),
+    };
+    this._scans.push(scanInfo);
+    return this._tx.scan(options, (key, isLastBeforeLimit, isLastEntry) => {
+      scanInfo.endInclusiveKey =
+        isLastEntry && !isLastBeforeLimit ? undefined : key;
```
arv: This does not look right. See other comment

src/btree/node.ts
```diff
@@ -232,6 +237,9 @@ export class DataNodeImpl extends NodeImpl<ReadonlyJSONValue> {
       limit > 0 && i < entries.length && entries[i][0].startsWith(prefix);
       limit--, i++
     ) {
+      if (onKey) {
+        onKey(entries[i][0], limit === 1, i === entries.length - 1);
```
arv: I do not understand why we have the third argument (`isLastEntry`) here? It only means that it is the last entry in the current chunk/DataNode. There might be more entries in the next chunk/DataNode.

Given a tree like this:

```
              InternalNode
      [['b', hash1], ['d', hash2]]
         /                    \
     DataNode              DataNode
[['a', 0, ['b', 1]]   [['c', 2, ['d', 3]]
````

The key `'b'` has no significance in `scan({})`

Comment (grgbkr) [reply src/btree/node.ts]: I see.  

I've removed this as it was only needed for the version that tried to optimize the case where the scan was not read to limit.  

In particular, if the scan ran out of entries before hitting limit, then we can't use the lastKey to rule out matches (because items can be added after the very last entry that should match the scan and our greater than lastKey, its very similar to why we agreed you can only use lastKey for prefix if the limit was hit).  

How would we determine if it is the very last entry (not in this chunk, but overall)?

File: src/transactions.ts (hunk 1 above)
Comment (grgbkr) [reply src/transactions.ts]: I've removed this as it was only needed for the version that tried to optimize the case where the scan was not read to limit.

**Comment by grgbkr** [comment]
@arv this is ready for review now.  Let me know if you can think of any other tests to write.

I also don't know too much about subscriptions/scan and indexes .  Are there any considerations around indexes we are missing?

File: src/transactions.ts [hunk 3]
```diff
@@ -10,6 +10,7 @@ import * as db from './db/mod';
 import * as sync from './sync/mod';
 import type {LogContext} from './logger';
 import type {Hash} from './hash';
+import type {ScanSubrscriptionInfo} from './subscriptions';
```
arv: typo

src/btree/node.ts
```diff
@@ -232,6 +233,9 @@ export class DataNodeImpl extends NodeImpl<ReadonlyJSONValue> {
       limit > 0 && i < entries.length && entries[i][0].startsWith(prefix);
       limit--, i++
     ) {
+      if (onKey) {
+        onKey(entries[i][0], limit === 1);
```
arv: I think we should check the limit here and call the callback in case limit is 1 (like I had it =P)

```ts
if (onLastKey && limit === 1) {
   onLastKey(entries[i][0]);
}
```

It feels unnecessary to call the function on every key, just to later check it was the last key or not.

src/subscriptions.test.ts
```diff
@@ -112,320 +111,424 @@ test('scanInfoMatchesKey', () => {
...
+          limit: 10,
+        },
+        inclusiveLimitKey: 'pac8',
+      },
+      '',
+      'pac1',
+    ),
+  ).to.be.true;
+
+  // Changed key matches prefix and equals nclusive limit key
```
arv: typo

arv (approved):
LGTM

src/subscriptions.ts
```diff
@@ -3,14 +3,19 @@ import type {ReadTransaction} from './transactions';
 import * as db from './db/mod';
 import type * as sync from './sync/mod';
 
+export type ScanSubrscriptionInfo = {
```
arv: typo here too

arv (approved):

src/subscriptions.ts
grgbkr: good catch, fixed

PR #672, opened 2021-11-08
feat: Update WriteSubRead benchmark to invalidate random keys (instead of a fixed set)

grgbkr:


src/range.ts
```diff
@@ -0,0 +1,25 @@
+export function range(
+  startInclusive: number,
+  endExclusive: number,
+  step?: number,
+): number[] {
```
arv: this makes me a bit uncomfortable. To me, `range` should return an iterator

web-test-runner.config.mjs
```diff
@@ -40,7 +41,7 @@ export default {
         'src/migrate/*.test.ts',
         'src/btree/*.test.ts',
       ],
-      browsers: [firefox, chromium, webkit],
+      browsers: [chromium],
```
arv: undo this. I assume it was just for debugging.

src/range.ts
```diff
@@ -0,0 +1,25 @@
...
+    return [];
+  }
+  if (
+    (endExclusive < startInclusive && definedStep > 0) ||
+    (endExclusive > startInclusive && definedStep < 0)
+  ) {
+    return [];
+  }
+  const length = Math.ceil(
+    Math.abs(endExclusive - startInclusive) / Math.abs(definedStep),
```
arv: One abs would have been enough:

```ts
Math.abs((endExclusive - startInclusive) / definedStep),
```

PR #676, opened 2021-11-09
refactor: Rework test config to avoid missing new tests in src/

grgbkr:
### Problem
New tests added to the `src/` dir are not run by `npm run test` unless the are explicitly added to `web-test-runner.config.mjs`.  This is unexpected as tests added to subdirs of src are automatically included, and so I was confused for sometime as to why a new test I added wasn't running.  

The reason for this is we don't want to run `src/worker.test.ts` with the other tests.

### Solution
Move `worker.test.ts` and `worker-test.ts` to a subdir `worker-tests`, and add `src/*.test.ts` to files config in `web-test-runner.config.mjs`.  Now new tests added to `src/` will automatically be included in `npm run test`.


arv (approved):
I like it

Issue #671, opened 2021-11-05
Implement Dueling Dags


arv:
Design Doc: https://dusty-lancer-e6c.notion.site/Simplified-DD1-1ed242a8c1094d9ca3734c46d65ffce4

This is a tracking bug with individual tasks. If a task needs further discussion please break it into its own issue.

- [x] Implement `dag.LazyStore` @grgbkr (7bc6106)
  - [x] Use LRU based on `getSizeOfValue`? @grgbkr (7bc6106)
- [x] integrate  `dag.LazyStore` @grgbkr  (b770a25)
- [x] `dag/clients.ts` containing `Client` type and client helpers @grgbkr  (https://github.com/rocicorp/replicache/commit/9533421da8c2dcd39a206bac26788f7e69e71ebc)

## Persist

See https://www.notion.so/MemDag-with-TempHash-e42058ee8f354e7e918c433cd92191fb?d=654c8bde3d1b4d4994792b65b7049352#3dae910d006646b0b3af44877a70a996

- [x] Make the hash function/chunk creation part of dag/db so that memdag and perdag can use different hashing strategies @arv
  - [x] Maybe also make the `assertNotTempHash` be `dagStore.isValidHash` @arv
  - [x] B+Tree cannot use isTempHash as a bit for mutable. @arv
- [x] Gather chunks from memdag @arv
- [x] Write chunks to perdag @arv
- [x] Compute hashes off the IDB transaction @arv
- [x] Fixup memdag @arv 
- [x] Glue it together @arv

## Heartbeats
- [x] Heartbeats @grgbkr  (https://github.com/rocicorp/replicache/commit/6b18ad07e3c3afd2978c8c37e0a57ae34c8d16c1)
- [x] Update heartbeat timestamp on persist @arv 
## Collecting Client-Side Client State
- [x] Collecting Client-Side Client State @grgbkr (https://github.com/rocicorp/replicache/commit/9a4a5100173ac577e167f5838c6b35d4a7b60cf1)
- ~~[ ] rocicorp/replicache#784~~ Punted to a follow up.
- ~~[ ] rocicorp/mono#121~~ Punted to a follow up. 

## Bootstrap
  - [x] Pick the "best" client @grgbkr (https://github.com/rocicorp/replicache/commit/981248bf0caefd60754db4a64eb59524250dda8b) 
[...]

PR #683, opened 2021-11-12
feat: Functionality for managing the ClientMap needed for Simplified Dueling Dags

grgbkr:
In the Simplified Dueling Dags design for Realtime Persistence, each tab is a `client` and has its own `perdag` - an instance of `dag.Store` backed by IDB.  All tabs' `perdag` instances are backed by the same IDB object store, thus they share physical storage. 

To keep track of each client's current `headHash` (and additional metadata such as heartbeatTimestampMS used for garbage collection of client perdags), a new `ClientMap` data structure is introduced.  The `ClientMap` is stored in a chunk in the `perdag` at the head `'clients'`.  This `ClientMap` chunk contains refs to each client's `headHash`.

This change implements helpers for reading and writing the `ClientMap`.     

Part of #671

### DIFF
--- a/src/json.ts
+++ b/src/json.ts
@@ -1,4 +1,4 @@
-import {throwInvalidType} from './asserts';
+import {assertObject, throwInvalidType} from './asserts';
 import {hasOwn} from './has-own';
 
 /** The values that can be represented in JSON */
@@ -178,12 +178,13 @@ export function assertJSONValue(v: unknown): asserts v is JSONValue {
       if (Array.isArray(v)) {
         return assertJSONArray(v);
       }
-      return assertJSONObject(v as Record<string, unknown>);
+      return assertJSONObject(v);
   }
   throwInvalidType(v, 'JSON value');
 }
 
-function assertJSONObject(v: Record<string, unknown>): asserts v is JSONObject {
+export function assertJSONObject(v: unknown): asserts v is JSONObject {
+  assertObject(v);
   for (const k in v) {
     if (hasOwn(v, k)) {
       const val = v[k];
@@ -196,6 +197,7 @@ function assertJSONObject(v: Record<string, unknown>): asserts v is JSONObject {
     }
   }
 }
+
 function assertJSONArray(v: unknown[]): asserts v is JSONValue[] {
   for (let i = 0; i < v.length; i++) {
     const val = v[i];
--- a/src/sync/clients.test.ts
+++ b/src/sync/clients.test.ts
@@ -1,0 +1,528 @@
+import {expect} from '@esm-bundle/chai';
+import {MemStore} from '../kv/mod';
+import * as dag from '../dag/mod';
+import {assertHash, hashOf, initHasher, newTempHash} from '../hash';
+import {getClient, getClients, setClient, setClients} from './clients';
+
+setup(async () => {
+  await initHasher();
+});
+
+test('getClients with no existing ClientMap in dag store', async () => {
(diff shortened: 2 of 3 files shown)


src/json.ts
```diff
@@ -178,12 +178,13 @@ export function assertJSONValue(v: unknown): asserts v is JSONValue {
...
       }
-      return assertJSONObject(v as Record<string, unknown>);
+      return assertJSONObject(v);
   }
   throwInvalidType(v, 'JSON value');
 }
 
-function assertJSONObject(v: Record<string, unknown>): asserts v is JSONObject {
+export function assertJSONObject(v: unknown): asserts v is JSONObject {
+  assertObject(v);
```
arv, line 187: It was done this way because we already checked that v was an object before.

I guess now that it got exported we do not know the callers. It seems like we could split into an internal function and an exported one.

src/sync/clients.ts
```diff
@@ -0,0 +1,90 @@
...
+import type {ReadonlyJSONValue} from '../json';
+import {assertNumber, assertObject} from '../asserts';
+import {hasOwn} from '../has-own';
+
+// TODO: Make ClientID an opaque type
+type ClientID = string;
+type ClientMap = Map<ClientID, Client>;
+
+type Client = {
+  // A UNIX timestamp in milliseconds updated by the client once a minute
```
arv, line 12: JSDoc helps VSCode.

src/sync/clients.ts
```diff
@@ -0,0 +1,90 @@
...
+  assertHash(headHash);
+}
+
+function chunkDataToClientMap(chunkData?: ReadonlyJSONValue): ClientMap {
+  assertObject(chunkData);
+  const clients: ClientMap = new Map();
+  for (const key in chunkData) {
+    if (hasOwn(chunkData, key)) {
+      const value = chunkData[key];
+      if (value !== undefined) {
```
arv, line 34: At some point we should try to fix this type to not allow undefined.

src/sync/clients.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function setClients(
+  clients: ClientMap,
+  dagWrite: dag.Write,
+): Promise<void> {
+  const chunkData = clientMapToChunkData(clients);
+  const chunk = dag.Chunk.new(
+    chunkData,
+    Array.from(clients.values()).map(client => client.headHash),
```
arv, line 74: Array.from takes a mapping function:

```js
Array.from(clients.values(), client => client.headHash),
```

src/sync/clients.ts
```diff
@@ -0,0 +1,90 @@
...
+  );
+  return Promise.all([
```
arv, line 76: I think I prefer using async function rather than using `then`

src/sync/clients.test.ts
```diff
@@ -0,0 +1,528 @@
...
+      },
+      client2: {
+        heartbeatTimestampMs: 3000,
+        headHash: hashOf('head of commit client2 is currently at'),
+      },
+    }),
+  );
+  await dagStore.withWrite(async (write: dag.Write) => {
+    await setClients(clientMap, write);
+    return write.commit();
```
arv, line 35:
```js
await write.commit();
```

arv: I know it is the same but clearer IMO

arv (commented):
LGTM

PR #686, opened 2021-11-13
Introduce poke

aboodman:


src/replicache.ts
```diff
@@ -808,6 +813,38 @@ export class Replicache<MD extends MutatorDefs = {}> {
     this._pullConnectionLoop.send();
   }
 
+  /**
+   * Applies an update from the server to Replicache.
+   * Throws an error if cookie does not match. Host should re-init.
```
arv, line 818: > Host should re-init.

Can you expand on this?

src/replicache.test.ts
```diff
@@ -1616,6 +1616,84 @@ testWithBothStores('HTTP status push', async () => {
   expect(okCalled).to.equal(true);
 });
 
+testWithBothStores('poke', async () => {
+  // Things to test:
```
arv, line 1620: is this a todo or is it explaining what this is testing? Please rephrase to make it clear

src/sync/pull.ts
```diff
@@ -136,10 +152,7 @@ export async function beginPull(
       (response.cookie ?? null) === baseCookie
     ) {
       const syncHead = emptyHash;
-      return {
-        httpRequestInfo,
-        syncHead,
-      };
+      return syncHead;
```
arv, line 155:
```js
return emptyHash;
```

src/sync/pull.ts
```diff
@@ -195,13 +208,7 @@ export async function beginPull(
...
     const commitHash = await dbWrite.commit(SYNC_HEAD_NAME);
 
-    return {
-      httpRequestInfo: {
-        httpStatusCode: 200,
-        errorMessage: '',
-      },
-      syncHead: commitHash,
-    };
+    return commitHash;
```
arv, line 211: inline

arv (approved):

src/replicache.ts
phritz: I think this should say *client* should re-init: the server has sent us an update on top of a cookie it thinks we have but we don't. The client disconnecting and re-registering ensures that the server knows the latest cookie the client has, because it is included in the registration request.

src/replicache.test.ts
phritz: These are TODOs but I'm still trying to understand the specific cases.

src/sync/pull.ts
phritz: Do you mean
```
return await dbWrite.commit(SYNC_HEAD_NAME);
```
?

phritz (commented):
Looks like I'm going to land this thing as soon as I understand it, but first a few clarifications.

src/replicache.ts
```diff
@@ -808,6 +813,38 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+   *
+   * @experimental - This method is under development and its semantics will change.
+   */
+  async poke(poke: Poke): Promise<void> {
+    await this._ready;
+    const clientID = await this._clientIDPromise;
+    const requestID = sync.newRequestID(clientID);
```
phritz, line 825: I feel like this doesn't make much sense anymore unless it was set by the server. The point of it was to be able to tie actions together across client and server, so as it is it's a little bit useless.

src/sync/pull.ts
```diff
@@ -101,29 +100,46 @@ export async function beginPull(
...
-      mainHeadPostPull,
-      dagRead,
-    );
-    if (baseSnapshot.chunk.hash !== baseSnapshotPostPull.chunk.hash) {
-      throw new Error('Overlapping syncs JSLogInfo');
+    const baseSnapshot = await db.Commit.baseSnapshot(mainHead, dagRead);
+    const [baseLastMutationID, baseCookie] =
+      db.Commit.snapshotMetaParts(baseSnapshot);
+
+    if (!deepEqual(expectedBaseCookie, baseCookie)) {
```
phritz, line 133: I think this condition is weaker than what we had before, which was a check that the hash of the chunk didn't change. This checks that the cookie didn't change. Is it the case that two snapshots are exactly identical if and only if their cookies are identical? 
- is it not possible for users to not use cookies, in which case this check always passes even if snapshot changes
- i think it *is* possible at least theoretically that the LMID changes without the cookie changing, which makes this check incorrect I would think

Not sure about other cases? The original intent was "did the base snapshot change" and here we are using "is the cookie the same", which I'm not sure is saying the same thing.

src/replicache.ts
aboodman: This comment is saying that the current semantics in this PR are that if this method throws, caller (the app) should re-connect to its server because the cookies have gotten out of sync.

This is temporary -- in the near future I imagine Replicache implementing poke as a first-class concept and in that case, there would be a new `connect()` callback analagous to the current `pusher()` and `puller()` callbacks that Replicache would call to get host to re-connect.

src/replicache.ts
aboodman: Agree.

phritz (commented):
OK I think I understand enough to make the tiny edits required to land this. One significant question below.

src/sync/pull.ts
aboodman: It's not the same. I am not 100% that the code as written makes sense for pull. But I think I am about to advocate for ditching pull from the protocol so perhaps it doesn't matter?

phritz:
Closing this in favor of https://github.com/rocicorp/replicache/pull/714

PR #688, opened 2021-11-15
feat: Implement heartbeats mechanism needed for Client state garbage collection for Simplified Dueling Dags

grgbkr:
Simplified Dueling Dags requires a mechanism for collecting the perdag state for Clients (i.e. tabs) which have been closed.

A Client (i.e. tab) that has been closed cannot reliably clean up its own state (due to crashes and force closes).  It is difficult for other Client (i.e. tabs) to determine if a tab has been closed and is gone for ever, or just has been frozen for a long time.  The approach taken here is to have each Client update a heartbeatTimestampMS once per minute while it is active.  Other Client's then collect a Client only if it hasn't been active for a very long time (current plan is 1 week).

A client's heartbeat time is also updated when its memdag is persisted to the perdag.  This way the "newest" client state is roughly the state of the client with the most recent heartbeat time, which is useful for determining which client state a new client should choose for bootstrapping. 

A timestamp is used (as opposed to a heartbeat counter) in order to support expiration periods much longer than a typical session (e.g. 7 days).

Part of #671

### DIFF
--- a/src/sync/heartbeat.ts
+++ b/src/sync/heartbeat.ts
@@ -1,0 +1,40 @@
+import type {ClientID} from './client-id';
+import type * as dag from '../dag/mod';
+import {getClient, setClient} from './clients';
+
+const HEARTBEAT_INTERVAL_MS = 60 * 1000;
+
+export function startHeartbeats(
+  clientID: ClientID,
+  dagStore: dag.Store,
+): () => void {
+  const intervalID = window.setInterval(async () => {
+    await dagStore.withWrite(async (write: dag.Write) => {
+      await writeHeartbeat(clientID, write);
+      await write.commit();
+    });
+  }, HEARTBEAT_INTERVAL_MS);
+  return () => {
+    window.clearInterval(intervalID);
+  };
+}
+
+export async function writeHeartbeat(
+  clientID: ClientID,
+  write: dag.Write,
+): Promise<void> {
+  const client = await getClient(clientID, write);
+  if (!client) {
+    // Should this be a more specific error so caller can detect and handle?
+    throw new Error('Cannot write heartbeat. Client with clientID not found');
+  }
+
+  await setClient(
+    clientID,
+    {
+      ...client,
+      heartbeatTimestampMs: Date.now(),
+    },
+    write,
+  );
+}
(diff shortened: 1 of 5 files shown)


src/sync/heartbeat.ts
```diff
@@ -0,0 +1,40 @@
...
+  };
+}
+
+export async function writeHeartbeat(
+  clientID: ClientID,
+  write: dag.Write,
+): Promise<void> {
+  const client = await getClient(clientID, write);
+  if (!client) {
+    // Should this be a more specific error so caller can detect and handle?
```
grgbkr, line 28: Haven't completely thought through all the places where missing Client (or client chunk) is going to arise and how we are going to handle it.

src/sync/heartbeat.ts
```diff
@@ -0,0 +1,40 @@
...
+    throw new Error('Cannot write heartbeat. Client with clientID not found');
+  }
+
+  await setClient(
+    clientID,
+    {
+      ...client,
```
arv: Can you mark the Client type as readonly so we do not accidentally mutate it?

src/sync/heartbeat.ts
arv: I agree that it might need some more thinking about... Just leave this comment and we can come back to it when more things are in place.

src/sync/heartbeat.ts
arv: I've been careful to always write these JSON objects in the same order as described in the typedef. The order matters when computing the hash. Things will not break but generally we want to ensure that we end up with the same hash for 2 json values that are deep equal.

In this specific case it is not going to matter because `heartbeatTimestampMs` is changing every time.

arv (approved):

src/sync/heartbeat.ts
grgbkr: done.  

the object construction details impacting hash is a nasty gotcha.

aboodman: Yeah! I wonder if we can put a lint pass in that they are alpha?

aboodman: (Note: hash mismatches don’t affect our correctness, only performance)

PR #689, opened 2021-11-15
feat: Implements Client state Garbage Collection for Simplified Dueling Dags

grgbkr:
Simplified Dueling Dags requires a mechanism for collecting the perdag state for Clients (i.e. tabs) which have been closed.

Every **five minutes**, each Client collects any Clients that haven't updated their heartbeat timestamp **for at least seven days**. 

Part of #671

### DIFF
--- a/src/sync/client-gc.test.ts
+++ b/src/sync/client-gc.test.ts
@@ -1,0 +1,189 @@
+import {expect} from '@esm-bundle/chai';
+import {SinonFakeTimers, useFakeTimers} from 'sinon';
+import {MemStore} from '../kv/mod';
+import * as dag from '../dag/mod';
+import {getClients, setClient, setClients} from './clients';
+import {hashOf, initHasher} from '../hash';
+import {initClientGC} from './client-gc';
+
+let clock: SinonFakeTimers;
+const START_TIME = 0;
+const SEVEN_DAYS_IN_MS = 7 * 24 * 60 * 60 * 1000;
+const FIVE_MINS_IN_MS = 5 * 60 * 1000;
+setup(async () => {
+  await initHasher();
+  clock = useFakeTimers(0);
+});
+
+teardown(() => {
+  clock.restore();
+});
+
+test('initClientGC starts 5 min interval that collects clients that have been inactive for > 7 days', async () => {
+  const dagStore = new dag.Store(new MemStore());
+  const client1 = {
+    heartbeatTimestampMs: START_TIME,
+    headHash: hashOf('head of commit client1 is currently at'),
+  };
+  const client2 = {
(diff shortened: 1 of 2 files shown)


src/sync/client-gc.ts
```diff
@@ -0,0 +1,32 @@
...
+      const now = Date.now();
+      for (const [id, client] of clients) {
+        if (
+          id !== clientID /* never collect ourself */ &&
+          now - client.heartbeatTimestampMs > CLIENT_MAX_INACTIVE_IN_MS
+        ) {
+          clients.delete(id);
+        }
+      }
+      await setClients(clients, write);
```
arv: Only write/commit if something changed?

src/sync/client-gc.test.ts
```diff
@@ -0,0 +1,189 @@
...
+  await dagStore.withRead(async (read: dag.Read) => {
+    const readClientMap = await getClients(read);
+    expect(readClientMap).to.deep.equal(clientMap);
+  });
+
+  clock.tick(SEVEN_DAYS_IN_MS + 1);
+
+  // client1 is not collected because it is the current client (despite being old enough to collect)
+  // client2 is collected because it is > 7 days inactive
+  // client3 is not collected because its < 7 days inactive (by 1 minute)
```
arv: its -> it's

arv: "its" is possessive, "it's" is short for "it is"

arv (approved):
Looking good

src/sync/client-gc.ts
grgbkr: done

src/sync/client-gc.test.ts
grgbkr: done

src/sync/client-gc.ts
```diff
@@ -0,0 +1,31 @@
...
+  const intervalID = window.setInterval(async () => {
+    await dagStore.withWrite(async (write: dag.Write) => {
+      const clients = await getClients(write);
+      const now = Date.now();
+      const clientsAfterGC = Array.from(clients).filter(
+        ([id, client]) =>
+          id === clientID /* never collect ourself */ ||
+          now - client.heartbeatTimestampMs <= CLIENT_MAX_INACTIVE_IN_MS,
+      );
+      if (clientsAfterGC.length !== clients.size) {
```
arv, line 21: You could have kept the old code and just add a flag.

```js
let changed = false;
...
changed = true;
...
if (changed) {
  ...
}
```

Anyway, the current code works and is not that much complicated

grgbkr: yes I considered that as well but decided I liked this functional programming version a tiny bit better.

PR #696, opened 2021-11-16
fix: Fix bug where pusher/puller/pushURL/pullURL set after construction are ignored if initially none set.

### LINKED ISSUE #685: Cannot change `pusher` or `puller` after initialization
I think this is a regression in 8.0, but I'm not sure:

```ts
const r = new Replicache();
r.pusher = ...
r.puller = ...
```

The pusher and puller impls won't get called because we create a `NullConnectionLoop` at startup and it is never changed. I think it might work OK if you put a no-op pusher/puller in there instead though.

grgbkr:
Also updates tests to cover these cases.

Fixes #685


arv (commented):
LGTM

PR #712, opened 2021-11-19
feat: Simplified Dueling Dags - Implement initing a new client including bootstraping from existing client state.

grgbkr:
Simplified Dueling Dags always creates a new Client for each new tab.  To enable fast startup of new tabs utilizing previous stored data Simplified Dueling Dags bootstraps new clients by forking an existing Client's state. 

When forking from another Client, the fork should be based on the existing Client's most recent base snapshot (which may not be its latest head).  This is necessary because pending mutations (LocalMutationCommits) cannot be forked as the last mutation id series is different per client.

It is important that the last mutation id for the new client be set to 0, since a replicache server implementation will start clients for which they do not have a last mutation id stored at last mutation id 0.  If the server receives a request from a client with a non-0 last mutation id, for which it does not have a last mutation id stored, it knows that it is unsafe for it to execute mutations form the client, as it could result in re-running mutations or otherwise failing to guarantee sequential execution of mutations.  This tells the server that this is an old client that it has GC'd (we need some way to signal this to the client so it can reset itself, see https://github.com/rocicorp/replicache/issues/335). 

When choosing a Client to bootstrap from, it is safe to pick any Client, but it is ideal to chose the Client with the most recent snapshot from the server.  Currently the age of snapshots is not stored, so this implementation uses a heuristic of choosing the base snapshot of the Client with the newest heartbeat timestamp. 

See larger design at https://www.notion.so/Simplified-DD1-1ed242a8c1094d9ca3734c46d65ffce4#64e4299105dd490a9ffbc6c9c771f5d2

Part of #671

### DIFF
--- a/src/db/commit.ts
+++ b/src/db/commit.ts
@@ -2,6 +2,7 @@ import type * as dag from '../dag/mod';
 import type {ReadonlyJSONValue} from '../json';
 import {assertJSONValue} from '../json';
 import {
+  assert,
   assertArray,
   assertNotNull,
   assertNotUndefined,
--- a/src/db/commit.ts
+++ b/src/db/commit.ts
@@ -2,7 +2,6 @@ import type * as dag from '../dag/mod';
 import type {ReadonlyJSONValue} from '../json';
 import {assertJSONValue} from '../json';
 import {
-  assert,
   assertArray,
   assertNotNull,
   assertNotUndefined,
--- a/src/sync/clients.ts
+++ b/src/sync/clients.ts
@@ -4,10 +4,13 @@ import type {ReadonlyJSONValue} from '../json';
 import {assertNumber, assertObject} from '../asserts';
 import {hasOwn} from '../has-own';
 import type {ClientID} from './client-id';
+import {uuid as makeUuid} from './uuid';
+import {Commit, newSnapshot} from '../db/commit';
+import {BTreeWrite} from '../btree/write';
 
 type ClientMap = Map<ClientID, Client>;
 
-type Client = {
+export type Client = {
   /**
    * A UNIX timestamp in milliseconds updated by the client once a minute
    * while it is active and everytime the client persists its state to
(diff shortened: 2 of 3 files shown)


src/sync/clients.ts
```diff
@@ -92,3 +95,55 @@ export async function setClient(
...
+      newClientCommit = newSnapshot(
+        dagWrite.createChunk,
+        bootstrapCommit.meta.basisHash,
+        0 /* lastMutationID */,
+        bootstrapCommit.meta.cookieJSON,
+        bootstrapCommit.valueHash,
+        bootstrapCommit.indexes,
+      );
+    } else {
+      // No existing snapshot to bootstrap from. Create empty snapshot.
```
grgbkr: This has the same effect as`initDB` in `db/write.ts`, but doesn't set the dag head `main` to the commits `hash`.  `initDB` uses the more general and complicated `db.Write#commitWithChangedKeys` code path.

I believe in Simplified Dueling Dags `db.Write` will only be used on the memdag, and not on the permdag.

grgbkr (commented):
@arv I wanted to get your initial feedback on this code before writing tests for it. 

What do you think?

src/db/commit.ts
```diff
@@ -98,7 +98,10 @@ export class Commit<M extends Meta = Meta> {
     return commits.filter(c => c.isLocal()) as Commit<LocalMeta>[];
   }
 
-  static async baseSnapshot(hash: Hash, dagRead: dag.Read): Promise<Commit> {
+  static async baseSnapshot(
+    hash: Hash,
+    dagRead: dag.Read,
+  ): Promise<Commit<SnapshotMeta>> {
```
arv: Thanks, I wonder if we should remove the default value for the type parameter?

src/sync/clients.ts
```diff
@@ -92,3 +95,55 @@ export async function setClient(
...
   return setClients(clients, dagWrite);
 }
+
+export async function initClient(
+  dagStore: dag.Store,
+): Promise<[ClientID, Client]> {
+  return dagStore.withWrite(async dagWrite => {
+    const clients = await getClients(dagWrite);
+    const newClientID = makeUuid();
+    const bootstrapClient = [...clients.values()].reduce((prev, curr) =>
```
arv: this is not hot code but you could have done this with a loop so you do not have to allocate that temp array.

src/sync/clients.ts
```diff
@@ -92,3 +95,55 @@ export async function setClient(
   clients.set(id, client);
   return setClients(clients, dagWrite);
 }
+
+export async function initClient(
+  dagStore: dag.Store,
```
arv: Maybe pass in dag.Write instead? It is not clear yet how this will be used so we will see if this needs its own transaction or not?

src/sync/clients.ts
```diff
@@ -92,3 +95,55 @@ export async function setClient(
...
+      );
+      // Copy the snapshot with one change: set last mutation id to 0.
+      newClientCommit = newSnapshot(
+        dagWrite.createChunk,
+        bootstrapCommit.meta.basisHash,
+        0 /* lastMutationID */,
+        bootstrapCommit.meta.cookieJSON,
+        bootstrapCommit.valueHash,
+        bootstrapCommit.indexes,
+      );
```
arv: Maybe add a `setLastMutationID` method instead?

src/sync/clients.ts
arv: I think you are right. The perdag does use db.Commit but only in the sense that it uses it to create chunks

arv (commented):
Seems good

src/db/commit.ts
grgbkr: I put up https://github.com/rocicorp/replicache/pull/715 for this.

src/sync/clients.ts
grgbkr: I'm going to just write lazy map/reduce/filter for iterator because IMHO functional codes intent is quicker to grok then declarative.

src/sync/clients.ts
grgbkr: good point, done.

src/sync/clients.ts
grgbkr: Not sure how to do this.  Commit and CommitData are both immutable.

grgbkr:
I still need to write tests for this.

src/sync/clients.ts
grgbkr: went ahead and made this a for/of until I manage to write some functional iterator helpers

grgbkr:
now there are tests!

arv (approved):
LGTM

PR #715, opened 2021-11-19
refactor: remove default value of Meta for Commit's  M type param to improve typing of Commit

grgbkr:
### Problem
Having a default value of Meta for Commit's M type param led to less specific typing in many places. 

### Solution
Remove the default and make typing more specific where possible.

### DIFF
--- a/src/db/commit.test.ts
+++ b/src/db/commit.test.ts
@@ -117,7 +117,7 @@ test('chain', async () => {
 });
 
 test('load roundtrip', async () => {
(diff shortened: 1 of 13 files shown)


arv (approved):
Not 100% sure this was the right call though. Up to you.

grgbkr:
I think it 

> Not 100% sure this was the right call though. Up to you.

I think it improved the typing specificity of ~5 functions.  It does require some more imports in other files, but on the whole I think better to be explicit.

PR #716, opened 2021-11-19
fix: Provide more details to logger when push or pull fails

### LINKED ISSUE #690: When pull fails it doesn't provide details
I hit this, as did a user. We set the `cause` of the exception, here:

https://github.com/rocicorp/replicache/blob/16d2597163a88c65f51d6632bf1b03fd294769cf/src/puller.ts#L101

But this doesn't show up in the console anywhere so it's hard to know what's going wrong. Maybe we should include the causes's message in our message?

Or maybe we don't even need this `PullError` abstraction anymore.

grgbkr:
### Problem
Push and pull error logging lacks sufficient detail to debug errors.

A pull failure currently logs to info (and a push error logs essentially the same):
`Pull returned: PullError: Failed to pull`

Not the most useful logging.  However, our `PushError` and `PullError` classes have a `cause?: Error` property with details on the underlying cause, it is just not logged.

### Solution
If the error is a `PushError` or `PullError`, log the cause.

Update log format to include stack traces for both the error, and cause.

Also update to use `error` instead of `info` logging.

Closes #690

Example error log message:
image

### DIFF
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -715,8 +715,10 @@ export class Replicache<MD extends MutatorDefs = {}> {
 
       if (e instanceof PushError || e instanceof PullError) {
         online = false;
+        this._logger.error?.(`${name} threw:\n`, e, '\nwith cause:\n', e.cause);
+      } else {
+        this._logger.error?.(`${name} threw:`, e);
       }
-      this._logger.info?.(`${name} returned: ${e}`);
       return false;
     } finally {
       if (this._online !== online) {
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -717,7 +717,7 @@ export class Replicache<MD extends MutatorDefs = {}> {
         online = false;
         this._logger.error?.(`${name} threw:\n`, e, '\nwith cause:\n', e.cause);
       } else {
-        this._logger.error?.(`${name} threw:`, e);
+        this._logger.error?.(`${name} threw:\n`, e);
       }
       return false;
     } finally {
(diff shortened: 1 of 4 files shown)


aboodman:
I think we chose `info` on purpose because from the pov of offline-first, a failed network request isn't an error. But maybe thinking about instant UI, it is!

Let's try it this way for awhile, I like it.

aboodman:
Just to be sure you actually saw visually that this gave the details in the console right?

aboodman (approved):

src/replicache.ts
```diff
@@ -715,8 +715,10 @@ export class Replicache<MD extends MutatorDefs = {}> {
 
       if (e instanceof PushError || e instanceof PullError) {
         online = false;
+        this._logger.error?.(`${name} threw:\n`, e, '\nwith cause:\n', e.cause);
```
arv: Needs a browser test too or we will print the cause twice

arv: This is with Firefox:

image

grgbkr: which browser does this I tested Chrome, Safari and Firefox on my mac and none of them printed the cause.

grgbkr:
> Just to be sure you actually saw visually that this gave the details in the console right?

Yep, and added a screenshot of this to the pull request.

src/replicache.ts
grgbkr: missed your comment above.  most be a newer version of firefox than I have.

grgbkr: browser checks would be ugly, to avoid this whole issue i just changed the field name from `cause` to `causedBy`, so we avoid the inconsistent behavior while browser adoption of the new standard is influx.

arv (commented):
OK

PR #719, opened 2021-11-22
fix: correct log levels for push and pull errors to follow style guidelines

grgbkr:
### Problem
Push and pull errors are being logged at level `error`, which violates our style guide for log levels: https://github.com/rocicorp/replicache/blob/main/CONTRIBUTING.md#style-general

### Solution
Update to use `info` level instead.

### DIFF
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -2790,7 +2790,7 @@ test('online', async () => {
     log.push(b);
   };
 
-  const consoleError = sinon.stub(console, 'error');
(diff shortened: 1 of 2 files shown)


phritz (approved):

arv:
This might affect #722. Why don't you land this first and I'll look at #722 after that

LGTM

grgbkr:
> This might affect #722. Why don't you land this first and I'll look at #722 after that
> 
> LGTM

thanks! landed

PR #724, opened 2021-11-24
 feat: Simplified Dueling Dags: Implement lazy kv store for memdag

grgbkr:
An implementation of kv.Store specifically for the needs of memdag in the Simplified Dueling Dags design. This is exactly like kv.MemStore except that the keys under c/ are lazy-loaded from the perdag's kv.Store These keys may be evicted at any time as long as they have been persisted to the perdag.

`LazyStore` lazily loads values from a base store and then caches them in memory.   In practice this base store will all be the IDBStore (i.e. the perdag in the Dueling Dags design).

Writes only manipulate the in memory cache (thus values must be persisted to the base store through a separate process).    

The memory caches size is limited to `cacheSizeLimit`, and values are evicted in an LRU fashion.  

`shouldBePinned` enables exempting some keys from eviction.  This mechanism should be used to prevent values not persisted to the base store from being lost.  Values whose keys are pinned cannot be evicted, and their sizes are not included in the current cache size.   In practice Dueling Dags will use an implementation of `shouldBePinned` which returns true for head keys (because heads are not persisted to the perdag), and for chunk keys containing temp hashes (because the temp hashes tell us they have not yet been persisted).   We use a pluggable predicate to avoid LazyStore having to know about head and chunk key structure.  LazyStore is in the kv store layer and should not know about chunks and heads which are concepts from the higher level dag layer.  

Once a value is persisted to the base store, it's temp key for which `shouldBePinned` is `true` can be deleted, and it can be re-put with a key for which `shouldBePinned` is false (making it subject to cache eviction, which is safe since it is now persisted).  In practice this will happen as part of the Dueling Dag [persist process](https://www.notion.so/Simplified-DD1-1ed242a8c1094d9ca3734c46d65ffce4#ffc02a501c864528936154de347c1869).  Once the persist process persists chunks with temp hashes to the perdag, it will put chunks with the permanent hashes into the LazyStore, and dag gc will delete the chunks with the temp hashes from the LazyStore.

See larger design at https://www.notion.so/Simplified-DD1-1ed242a8c1094d9ca3734c46d65ffce4#25008ba7094a472c98259b4c0f380301

Part of #671


src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+    this._release();
+    this._closed = true;
+  }
+
+  get closed(): boolean {
+    return this._closed;
+  }
+
+  async has(key: string): Promise<boolean> {
+    return this._cache.has(key) || (await this.getBaseRead()).has(key);
```
arv: I think we should cache the value here too

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (v === undefined) {
+      v = await (await this.getBaseRead()).get(key);
+      if (v !== undefined) {
+        this._cache.set(key, v);
+      }
+    }
+    return v;
+  }
+
+  private async getBaseRead(): Promise<Read> {
```
arv: leading underscore for private methods.

Can you fix the eslint config to catch this?

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+  get(key: string): Value | undefined {
+    if (this._shouldBePinned(key)) {
+      return this._pinned.get(key);
+    }
+    const cacheEntry = this._cacheEntries.get(key);
+    if (cacheEntry) {
+      this._cacheEntries.delete(key);
+      this._cacheEntries.set(key, cacheEntry);
+    }
+    return cacheEntry && cacheEntry.value;
```
arv:
```suggestion
    return cacheEntry?.value;
```

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  set(key: string, value: Value): void {
+    if (this._shouldBePinned(key)) {
+      this._pinned.set(key, value);
+      return;
+    }
+    const oldValue = this._cacheEntries.get(key);
+    if (oldValue) {
+      this._size -= getSizeOfValue(oldValue);
```
arv: don't you already have the size in the CacheEntry?

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+    const valueSize = getSizeOfValue(value);
+    this._size += valueSize;
+    this._cacheEntries.delete(key);
+    this._cacheEntries.set(key, {value, size: valueSize});
+
+    let curr = this._cacheEntries.entries().next();
```
arv: why not for of loop?

```js
for (const entry of this._cacheEntries) {
  if (this._size <= this._cacheSizeLimit) {
    break;
  }
  const [keyToEvict, cacheEntryToEvict] = entry;
  this._size -= cacheEntryToEvict.size;
  this._cacheEntries.delete(keyToEvict);
}
```

arv (changes requested):
This is a good start

src/kv/lazy-store.ts
grgbkr: yes, great catch, missed this in the refactor.  I refactored to cache the size.  So went from a Map<string, Value> to Map<string, CacheEntry>.  TypeScript didn't help so much with this refactor since CacheEntry types as Value.

src/kv/lazy-store.ts
grgbkr: i wasnt sure how for/of or Iterator in general behaved when you modify the underlying collection.  I'll give this a try, and if it works, agree it is better.

grgbkr (commented):
Thanks for the quick initial review.

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  constructor(
+    cacheSizeLimit: number,
+    shouldBePinned: (key: string) => boolean,
+  ) {
+    this._cacheSizeLimit = cacheSizeLimit;
+    this._shouldBePinned = shouldBePinned;
+  }
+
+  has(key: string): boolean {
```
aboodman: It seems like this needs to check `_shouldBePinned` too?

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
...
+class Cache {
+  private readonly _cacheSizeLimit: number;
+  private readonly _shouldBePinned: (key: string) => boolean;
+  private readonly _pinned = new Map<string, Value>();
+  private readonly _cacheEntries = new Map<string, CacheEntry>();
+  private _size = 0;
+
+  constructor(
+    cacheSizeLimit: number,
+    shouldBePinned: (key: string) => boolean,
```
aboodman, line 171: Is the idea that `shouldBePinned` will return true if the key is a head? Or if the key refers to a temp hash? Or both? Some comments here would be helpful (realize this is draft and you and Erik probably talked about this).

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,202 @@
+import {RWLock} from '../rw-lock';
+import type {Read, Store, Value, Write} from './store';
+import {deleteSentinel, WriteImplBase} from './write-impl-base';
+import {getSizeOfValue} from '../get-size-of-value';
+
+export class LazyStore implements Store {
```
aboodman, line 5: Would `MemStore` go away? It seems like `LazyStore` is the same as `MemStore` if `shouldBePinned` always returns true.

aboodman: (if so maybe the right name for this thing is `MemStore` :-)).

aboodman (commented):
I like where this is going. Clean factoring with the separate `Cache` class.

src/kv/lazy-store.ts
grgbkr: Exactly, shouldBePinned will be true for head keys and chunk keys with temp hashes. 

The tricky thing here is the abstraction layering.  With the current layering, this layer of abstraction should not know about chunks and heads etc... its just a key value store.  That is why I opted for a `shouldBePinned` function to be passed in rather than just explicitly checking for head keys/temp hashes.  For the same reason I'm not sure if we should put comments here about chunks keys/temp hashes. 

In this case though keeping the layer "clean" makes the design a little less clear.

src/kv/lazy-store.ts
grgbkr: done

src/kv/lazy-store.ts
grgbkr: Yes MemStore goes away.  I imagined it went away in a future change, but you are right, I could replace it with a LazyStore with shouldBePinned => true.  I like that.  I'll make it part of this change.

src/kv/lazy-store.ts
grgbkr: good catch. done

src/kv/lazy-store.ts
arv: I think a comment given our intended use case is helpful even when the solution is more generic

src/kv/lazy-store.ts
grgbkr: Actually I am going to wait till a later integration change to get rid of MemStore.  LazyStore needs a base Store, and I don't want to make the change to always create an IDBStore as part of this change.  At that time, we can also rename LazyStore to MemStore.

arv: It seems like MemStore can be removed and we can use kv.TestStore in all the tests

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,235 @@
...
+   * Store which lazily loads values from a base store and then caches them in memory.
+   *
+   * Writes only manipulate the in memory cache (thus values must be persisted to the
+   * base store through a separate process).
+   *
+   * The memory caches size is limited to `cacheSizeLimit`, and values are evicted in an LRU
+   * fashion.  `shouldBePinned` enables exempting some keys from eviction.  This mechanism
+   * should be used to prevent values not persisted to the base store from being lost.
+   * Values whose keys are pinned cannot be evicted, and their sizes are not included in the current
+   * cache size. Once a value is persisted to the base store, it's temp key for which `shouldBePinned`
```
arv:
```suggestion
   * cache size. Once a value is persisted to the base store, its temp key for which `shouldBePinned`
```

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,235 @@
...
+   */
+  constructor(
+    baseStore: Store,
+    cacheSizeLimit: number,
+    shouldBePinned: (key: string) => boolean,
+    getSizeOfValue = defaultGetSizeOfValue,
+  ) {
+    this._baseStore = baseStore;
+    this._cache = new Cache(
+      Math.max(cacheSizeLimit, 0),
```
arv: This does not seem needed. We are the consumers of this and we are not going to pass in negative numbers.

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,235 @@
...
+  get closed(): boolean {
+    return this._baseStore.closed;
+  }
+}
+
+class ReadImpl implements Read {
+  private readonly _cache: Cache;
+  private readonly _release: () => void;
+  private readonly _base: Store;
+  private _baseRead: Read | undefined;
```
arv: This needs an initializer.

I'm surprised TS is not complaining.  https://www.typescriptlang.org/tsconfig#strictPropertyInitialization

> this.address is declared as potentially undefined which means it does not have to be set.

super lame

arv: It is important to initialize all properties because adding a property later changes the hidden class (vm implementation detail) which leads to code going from monomorphic to multimorphic.

src/kv/lazy-store.ts
```diff
@@ -0,0 +1,235 @@
...
+    if (valueSize > this._cacheSizeLimit) {
+      // this value cannot be cached due to its size
+      // don't evict other entries to try to make room for it
+      return;
+    }
+    this._size += valueSize;
+    this._cacheEntries.set(key, {value, size: valueSize});
+
+    for (const entry of this._cacheEntries) {
+      if (this._size <= this._cacheSizeLimit) {
```
arv, line 214: My gut feeling is that it would be better to check this twice (once before the loop and once in the loop) so that we do not need to create the iterator in the common case but maybe it is too much of a premature optimization?

src/kv/lazy-store.test.ts
```diff
@@ -0,0 +1,639 @@
...
+
+runAll(
+  'lazystore everything cached (large cache limit, nothing pinned)',
+  () => new LazyStore(new TestMemStore(), 10000, _ => false),
+);
+
+const DEFAULT_VALUE_SIZE = 100;
+function getSizeOfValueForTest(value: Value): number {
+  if (value && typeof value === 'object' && !Array.isArray(value)) {
+    const {size} = value as Record<string, Value>;
```
arv:
```suggestion
    const {size} = value as {size?: Value};
```

src/kv/lazy-store.test.ts
```diff
@@ -0,0 +1,639 @@
...
+  await baseStore.withWrite(async write => {
+    await write.put('inBase1', 'inBase1Value');
+    await write.commit();
+  });
+  await lazyStore.withWrite(async write => {
+    await write.del('inBase1');
+    await write.commit();
+  });
+  await baseStore.withRead(async read => {
+    expect(await read.get('notPinned1')).to.be.undefined;
```
arv: Is this test testing the right thing?

I believe it should be:

```ts
test('del does not write through to base store', async () => {
  const {baseStore, lazyStore} = createLazyStoreForTest();
  await baseStore.withWrite(async write => {
    await write.put('inBase1', 'inBase1Value');
    await write.commit();
  });
  await lazyStore.withWrite(async write => {
    await write.del('inBase1');
    await write.commit();
  });
  await lazyStore.withRead(async read => {
    expect(await read.get('inBase1')).to.be.undefined;
  });
  await baseStore.withRead(async read => {
    expect(await read.get('inBase1')).to.equal('inBase1Value');
  });
});
```

src/kv/lazy-store.test.ts
```diff
@@ -0,0 +1,639 @@
...
+    await write.put('pinnedInBase1', 'pinnedInBase1V2Value');
+    await write.commit();
+  });
+  await lazyStore.withRead(async read => {
+    // value of pinnedInBase1 is from pinned cache, instead of new 'inBase1V2Value'
+    // currently in base store
+    expect(await read.get('pinnedInBase1')).to.equal('pinnedInBase1Value');
+    expect(await read.get('pinnedInBase2')).to.equal('pinnedInBase2Value');
+  });
+});
```
arv, line 142: to really test this you could clear the cache

src/kv/lazy-store.test.ts
```diff
@@ -0,0 +1,639 @@
...
+  });
+  await lazyStore.withWrite(async write => {
+    expect(await write.get('inBase1')).to.equal('inBase1Value');
+    expect(await write.get('inBase2')).to.equal('inBase2Value');
+    await write.del('inBase1');
+    await write.del('inBase2');
+    await write.commit();
+  });
+  await lazyStore.withRead(async read => {
+    // new values loaded from base store
```
arv, line 226: Is this the right behavior?

I was uncertain about this earlier and still am.

Trying to think through potential problems:
- Data chunks should be fine because those are immutable
- GC
  - Meta (aka refs key) chunks are not written when the refs are `[]`. Should be fine since these never change.
  - RefCount chunks are deleted when the count gets to 0
    - ~~If the count goes to 0 in the memdag it is deleted there, but if read again we will get the old count from the perdag.~~
    - https://github.com/rocicorp/replicache/blob/main/src/dag/write.ts#L157
    - https://github.com/rocicorp/replicache/blob/main/src/dag/write.ts#L138-L139

## BUT

GC happens at commit time so I think there is no problem after all.

src/kv/lazy-store.ts
grgbkr: Yes.  kv.TestStore is built on top of MemStore, so we'd need to merge MemStore into TestStore.

arv (changes requested):
Let's convince ourselves that the delete behavior is what we want.

src/kv/lazy-store.ts
grgbkr: done

src/kv/lazy-store.ts
grgbkr: done.  i knew about this performance detail at one time but had forgotten.  ill watch out going forward.

src/kv/lazy-store.ts
grgbkr: its a simple enough optimization. done.

src/kv/lazy-store.test.ts
grgbkr: done

src/kv/lazy-store.test.ts
grgbkr: good catch.  looks like a copy-pasta mistake.

rewrote to

```ts
test('del does not write through to base store', async () => {
  const {baseStore, lazyStore} = createLazyStoreForTest();
  await baseStore.withWrite(async write => {
    await write.put('inBase1', 'inBase1Value');
    await write.put('inBase2', 'inBase2Value');
    await write.commit();
  });
  await lazyStore.withRead(async read => {
    // load one of keys into the lazy store cache
    expect(await read.get('inBase1')).to.equal('inBase1Value');
  });
  await lazyStore.withWrite(async write => {
    // delete both keys (cached and not-cached) from lazy store
    await write.del('inBase1');
    await write.del('inBase2');
    await write.commit();
  });
  await baseStore.withRead(async read => {
    // neither delete impacts baseStore
    expect(await read.get('inBase1')).to.equal('inBase1Value');
    expect(await read.get('inBase1')).to.equal('inBase1Value');
  });
});
```

src/kv/lazy-store.ts
grgbkr: done

grgbkr:
Abandoning in favor of lazy dag store approach see https://github.com/rocicorp/replicache/pull/758

PR #752, opened 2021-12-06
feat: Move all client map updating to retrying updateClients pattern that enables using async native hashing outside of perdag transaction 

### LINKED ISSUE #735: Use a retry loop for setClients
After #671 we can (almost) remove the wasm hash function. The only remaining case is when we write the clients on the perdag.

When reading and changing the clients we need to compute the new hash. But using the native (`crypto.subtle.digest`) hash function is async which causes IDB transactions to auto commit.

Instead of relying on the synchronous wasm hash function we can retry until successful using the async native hash. Something along the lines of: 

```ts
async function updateClient(
  clientID: ClientID,
  headHash: Hash,
  dagStore: dag.Store,
): Promise<void> {
  const clients: ClientMap = await dagStore.withRead(dagRead =>
    getClients(dagRead),
  );

  const updatedClient: Client = {
    heartbeatTimestampMs: Date.now(),
    headHash,
  };
  const updatedClients = new Map(clients);
  updatedClients.set(clientID, updatedClient);
  const hash = await hashOfClients(updatedClients);

  const ok = await dagStore.withWrite(async dagWrite => {
    const clients2 = await getClients(dagWrite);
    if (!deepEqual(Object.fromEntries(clients), Object.fromEntries(clients2))) {
      return false;
    }
    clients2.set(clientID, updatedClient);
    await setClientsWithHash(updatedClients, hash, dagWrite);
    await dagWrite.commit();
    return true;
  });

  if (!ok) {
    // Clients changed between read and write transactions.
    return updateClient(clientID, headHash, dagStore);
  }
}
```

### LINKED ISSUE #743: If this tab crashes or is frozen before updateClients completes this temp head will never be cleaned up.
If this tab crashes or is frozen before updateClients completes this temp head will never be cleaned up.

We should update client-gc.ts, so that when it gc's a client it also cleans up any lingering temp heads for that client.

_Originally posted by @grgbkr in https://github.com/rocicorp/replicache/pull/740#discussion_r761362046_

grgbkr:
### Problem
For Simplified Dueling Dags we want to allow using an async native hash
function. That means that the hash of a chunk has to be computed
outside the DAG transaction (because of IDB's auto commit bug/feature).

This mostly works well on the perdag because it gets it's chunks from
the memdag using the persist function which allows us to precompute all
the hashes; **except for the hash of the clients map**.

### Solution
To not require a sync hash function we instead precompute the hash of
the clients map outside the DAG transaction and then write it in the tx.
However, by doing this there is a small chance that the clients map
changed since we mutated it and computed the hash for it. If it did
change we now retry the update clients function with the new up to date
clients map.

Fixes #735
Fixes #743

### DIFF
--- a/src/persist/client-gc.ts
+++ b/src/persist/client-gc.ts
@@ -5,9 +5,9 @@ import {ClientMap, noUpdates, updateClients} from './clients';
 const CLIENT_MAX_INACTIVE_IN_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
 const GC_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
 
-let _latestGCUpdate: Promise<ClientMap> | undefined;
+let latestGCUpdate: Promise<ClientMap> | undefined;
 export function getLatestGCUpdate(): Promise<ClientMap> | undefined {
-  return _latestGCUpdate;
+  return latestGCUpdate;
 }
 
 export function initClientGC(
@@ -15,7 +15,7 @@ export function initClientGC(
   dagStore: dag.Store,
 ): () => void {
   const intervalID = window.setInterval(() => {
-    _latestGCUpdate = updateClients(clients => {
+    latestGCUpdate = updateClients(clients => {
       const now = Date.now();
       const clientsAfterGC = Array.from(clients).filter(
         ([id, client]) =>
--- a/src/persist/client-gc.ts
+++ b/src/persist/client-gc.ts
@@ -18,11 +18,11 @@ export function initClientGC(
           now - client.heartbeatTimestampMs <= CLIENT_MAX_INACTIVE_IN_MS,
       );
       if (clientsAfterGC.length === clients.size) {
-        return Promise.resolve(noUpdates);
+        return noUpdates;
       }
-      return Promise.resolve({
+      return {
         clients: new Map(clientsAfterGC),
-      });
+      };
     }, dagStore);
   }, GC_INTERVAL_MS);
 
--- a/src/persist/client-gc.ts
+++ b/src/persist/client-gc.ts
@@ -1,6 +1,6 @@
 import type {ClientID} from '../sync/client-id';
 import type * as dag from '../dag/mod';
-import {getClients, setClients} from './clients';
+import {noUpdates, updateClients} from './clients';
 
 const CLIENT_MAX_INACTIVE_IN_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
 const GC_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
@@ -10,19 +10,20 @@ export function initClientGC(
   dagStore: dag.Store,
 ): () => void {
   const intervalID = window.setInterval(async () => {
-    await dagStore.withWrite(async (write: dag.Write) => {
-      const clients = await getClients(write);
+    await updateClients(clients => {
       const now = Date.now();
       const clientsAfterGC = Array.from(clients).filter(
         ([id, client]) =>
           id === clientID /* never collect ourself */ ||
           now - client.heartbeatTimestampMs <= CLIENT_MAX_INACTIVE_IN_MS,
       );
-      if (clientsAfterGC.length !== clients.size) {
-        await setClients(new Map(clientsAfterGC), write);
-        await write.commit();
+      if (clientsAfterGC.length === clients.size) {
+        return Promise.resolve(noUpdates);
       }
-    });
+      return Promise.resolve({
+        clients: new Map(clientsAfterGC),
+      });
+    }, dagStore);
   }, GC_INTERVAL_MS);
 
   return () => {
--- a/src/persist/client-gc.ts
+++ b/src/persist/client-gc.ts
@@ -1,16 +1,21 @@
 import type {ClientID} from '../sync/client-id';
 import type * as dag from '../dag/mod';
-import {noUpdates, updateClients} from './clients';
+import {ClientMap, noUpdates, updateClients} from './clients';
 
 const CLIENT_MAX_INACTIVE_IN_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
 const GC_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
+
+let _latestGCUpdate: Promise<ClientMap> | undefined;
+export function getLatestGCUpdate(): Promise<ClientMap> | undefined {
+  return _latestGCUpdate;
+}
 
 export function initClientGC(
   clientID: ClientID,
   dagStore: dag.Store,
 ): () => void {
-  const intervalID = window.setInterval(async () => {
-    await updateClients(clients => {
+  const intervalID = window.setInterval(() => {
+    _latestGCUpdate = updateClients(clients => {
       const now = Date.now();
       const clientsAfterGC = Array.from(clients).filter(
         ([id, client]) =>
@@ -25,7 +30,6 @@ export function initClientGC(
       };
     }, dagStore);
   }, GC_INTERVAL_MS);
-
   return () => {
     window.clearInterval(intervalID);
   };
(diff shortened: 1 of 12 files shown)


src/persist/heartbeat.test.ts
```diff
@@ -33,10 +34,7 @@ test('startHeartbeats starts interval that writes heartbeat each minute', async
       client2,
     }),
   );
-  await dagStore.withWrite(async (write: dag.Write) => {
-    await setClients(clientMap, write);
-    return write.commit();
-  });
+  await setClients(clientMap, dagStore);
```
grgbkr, line 48: This test and similar ones for GC are now failing because the updating of the client state in the dagstore is no longer synchronous on the clock.tick(ONE_MIN_IN_MS); (i.e. it is not done when the setInterval callback completes).  I'm not sure of a clean way to test this.

src/persist/clients.ts
```diff
@@ -72,83 +78,151 @@ export async function getClient(
...
+          bootstrapCommit.meta.basisHash,
+          0 /* lastMutationID */,
+          bootstrapCommit.meta.cookieJSON,
+          bootstrapCommit.valueHash,
+          bootstrapCommit.indexes,
+        );
+      } else {
+        // No existing snapshot to bootstrap from. Create empty snapshot.
+        // TODO This empty btree write is a problem...
+        //const emptyBTreeHash = await new BTreeWrite(dagWrite).flush();
```
grgbkr: This is the one significant difficulty remaining.  I need some way of getting the hash of the empty btree, and the chunks for the empty btree to return as chunksToPut.

src/persist/clients.ts
```diff
@@ -72,83 +78,151 @@ export async function getClient(
...
+): Promise<ClientMap> {
+  const [clients, clientsHash] = await dagStore.withRead(async dagRead => {
+    const clients = await getClients(dagRead);
+    const clientsHead = await dagRead.getHead(CLIENTS_HEAD);
+    return [clients, clientsHead];
+  });
+  return _updateClients(update, clients, clientsHash, dagStore);
+}
+
+export async function _updateClients(
```
arv: I don't like the name here ... maybe just updateClientsInternal?

src/persist/clients.ts
```diff
@@ -72,83 +78,151 @@ export async function getClient(
...
+  {clients: ClientMap; chunksToPut?: Iterable<dag.Chunk>} | NoUpdates
+>;
+
+export async function updateClients(
+  update: ClientsUpdate,
+  dagStore: dag.Store,
+): Promise<ClientMap> {
+  const [clients, clientsHash] = await dagStore.withRead(async dagRead => {
+    const clients = await getClients(dagRead);
+    const clientsHead = await dagRead.getHead(CLIENTS_HEAD);
```
arv: This does the head request twice

src/persist/clients.ts
```diff
@@ -72,83 +78,151 @@ export async function getClient(
...
+      await nativeHashOf(newClientCommitData),
+      newClientCommitData,
+      getRefs(newClientCommitData),
+    );
+    return {
+      clients: new Map(clients).set(newClientID, {
+        heartbeatTimestampMs: Date.now(),
+        headHash: newClientCommitChunk.hash,
+      }),
+      chunksToPut: [newClientCommitChunk],
```
arv: I like the factoring. `chunksToPut` is much cleaner than my hack using a temp head

src/persist/clients.ts
```diff
@@ -72,83 +78,151 @@ export async function getClient(
...
     );
+    const chunksToPutPromises: Promise<void>[] = [];
+    if (chunksToPut) {
+      for (const chunk of chunksToPut) {
+        chunksToPutPromises.push(dagWrite.putChunk(chunk));
+      }
+    }
+    await Promise.all([
+      ...chunksToPutPromises,
+      dagWrite.putChunk(updateClientsChunk),
```
arv, line 229: or just push to `chunksToPutPromises`

src/persist/heartbeat.test.ts
arv: isn't this what `await clock.tickAsync` etc is for? I use them in replicache.test.ts

src/persist/persist.ts
```diff
@@ -81,18 +81,14 @@ async function writeFixedChunks(
...
-      headHash: mainHeadHash,
-    };
-    // We need to set a head here or the chunks will be GC'd
-    ps.push(setClient(clientID, client, dagWrite));
-    await Promise.all(ps);
-    await dagWrite.commit();
-  });
+  const chunksToPut = fixedChunks.values();
+  await updateClients(clients => {
+    return Promise.resolve({
```
arv: or let the function return `MaybePromise<T>`

grgbkr: this is a neat one.  thanks for the suggestion.

src/persist/clients.ts
grgbkr: done

src/persist/clients.ts
```diff
@@ -110,31 +114,36 @@ export async function initClient(
...
-          null /* cookie */,
-          'emptybtreehash' as unknown as Hash,
-          [] /* indexes */,
-        );
-      }
-    });
+      });
+    } else {
+      // No existing snapshot to bootstrap from. Create empty snapshot.
+      // TODO This empty btree write is a problem...
```
arv: remove todo

src/persist/clients.ts
```diff
@@ -165,10 +174,10 @@ export async function updateClients(
     const clientsHead = await dagRead.getHead(CLIENTS_HEAD);
     return [clients, clientsHead];
   });
-  return _updateClients(update, clients, clientsHash, dagStore);
+  return updateClientsInternal(update, clients, clientsHash, dagStore);
 }
 
-export async function _updateClients(
+async function updateClientsInternal(
```
arv, line 186: rename usage too (`F2` is your friend)

arv (changes requested):

src/persist/heartbeat.test.ts
grgbkr: That didn't do it.  Seems tickAsync is for promises before the timers, not created by the timers.

https://sinonjs.org/releases/latest/fake-timers/
`The nextAsync() will also break the event loop, allowing any scheduled promise callbacks to execute before running the timers.`. 

I fixed this by exposing the promise to wait on.  Don't completely love it, but far cleaner than awaiting a setTimeout for X.

src/persist/clients.ts
grgbkr: fixed with Erik's approach.

src/persist/clients.ts
grgbkr: thanks.

src/persist/clients.ts
grgbkr: refactored to avoid the extra request

src/persist/clients.ts
grgbkr: i wanted chunksToPutPromises to correspond to the chunksToPut argument. so I think this is a bit clearer.

src/persist/clients.ts
grgbkr: done

src/persist/clients.ts
grgbkr: done

src/persist/client-gc.ts
```diff
@@ -1,30 +1,35 @@
 import type {ClientID} from '../sync/client-id';
 import type * as dag from '../dag/mod';
-import {getClients, setClients} from './clients';
+import {ClientMap, noUpdates, updateClients} from './clients';
 
 const CLIENT_MAX_INACTIVE_IN_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
 const GC_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
 
+let _latestGCUpdate: Promise<ClientMap> | undefined;
```
arv: My rule is that underscore is for private/protected properties/methods. Not for variable bindings

arv (approved):
Just some naming nits

src/persist/client-gc.ts
grgbkr: cool yeah I'm mixing up class private vs module private (i.e. not exported).

PR #758, opened 2021-12-08
 lazy dag store exploration

grgbkr:


src/dag/gc.ts
```diff
@@ -0,0 +1,90 @@
...
+export class GarbargeCollector {
+  private readonly _setRefCount: (hash: Hash, count: number) => Promise<void>;
+  private readonly _getRefCount: (hash: Hash) => Promise<number>;
+  private readonly _deleteChunk: (hash: Hash) => Promise<void>;
+  private readonly _getRefs: (
+    hash: Hash,
+  ) => Promise<readonly Hash[] | undefined>;
+
+  constructor(
+    setRefCount: (hash: Hash, count: number) => Promise<void>,
```
grgbkr, line 17: this is a lot of function that take a hash, passing them in the right order is a bit finicky, change to accept a single object implementing an interface.

src/dag/store-impl.ts
```diff
@@ -0,0 +1,50 @@
+import type {Hash} from '../hash';
+import type * as kv from '../kv/mod';
+import type {ChunkHasher} from './chunk';
+import {ReadImpl} from './read-impl';
+import type {Read, Write} from './store';
+import {WriteImpl} from './write-impl';
+
+export class StoreImpl {
```
grgbkr, line 8: Consider folding the read-impl and write-impl files into this file.

src/dag/lazy-store.ts
```diff
@@ -0,0 +1,350 @@
...
+    if (this._size <= this._cacheSizeLimit) {
+      return;
+    }
+    for (const entry of this._cacheEntries) {
+      if (this._size <= this._cacheSizeLimit) {
+        break;
+      }
+      const [keyToEvict, cacheEntryToEvict] = entry;
+      this._size -= cacheEntryToEvict.size;
+      this._cacheEntries.delete(keyToEvict);
```
grgbkr: need to update ref counts when we evict something from the cache.

src/dag/lazy-store.ts
```diff
@@ -0,0 +1,350 @@
...
+    return (await this.getChunk(hash)) !== undefined;
+  }
+
+  async getChunk(hash: Hash): Promise<Chunk | undefined> {
+    if (isTempHash(hash)) {
+      return this._tempChunks.get(hash);
+    }
+    let chunk = this._sourceChunksCache.get(hash);
+    if (chunk === undefined) {
+      chunk = await (await this._getSourceRead()).getChunk(hash);
```
grgbkr, line 152: need to update ref counts when we load a chunk from source

src/dag/lazy-store.ts
```diff
@@ -0,0 +1,350 @@
...
+      v.new = hash;
+    }
+  }
+
+  override async hasChunk(hash: Hash): Promise<boolean> {
+    return this._pendingChunks.get(hash) !== undefined || super.hasChunk(hash);
+  }
+
+  async getChunk(hash: Hash): Promise<Chunk | undefined> {
+    if (this._pendingChunks.has(hash)) {
```
arv: Do one hash lookup instead of two

src/dag/lazy-store.ts
```diff
@@ -0,0 +1,350 @@
...
+import type * as kv from '../kv/mod';
+import {RWLock} from '../rw-lock';
+import {Chunk, ChunkHasher, createChunk} from './chunk';
+import type {Store, Read, Write} from './store';
+import {getSizeOfValue as defaultGetSizeOfValue} from '../get-size-of-value';
+import type {ReadonlyJSONValue} from '../mod';
+import {GarbargeCollector, HeadChange} from './gc';
+
+export class LazyStore implements Store {
+  private readonly _rwLock = new RWLock();
```
aboodman, line 32: I think it's nice to add a comment saying what state/fields a lock is protecting.

aboodman: Side note: it would be neat to have a kind of lock that self-documents that, like:

```
RwLock<Stuff>
```
And then when you do `withLock` you receive the stuff as a param.

src/dag/lazy-store.ts
```diff
@@ -0,0 +1,350 @@
...
+    const headChange = this._pendingHeadChanges.get(name);
+    if (headChange) {
+      return headChange.new;
+    }
+    return super.getHead(name);
+  }
+
+  async commit(): Promise<void> {
+    const pendingRefCountUpdates = new Map<Hash, number | undefined>();
+    const garbargeCollector = new GarbargeCollector(
```
aboodman, line 268: Nit: Find this hard to read without names for the callbacks. Consider defining a `GarbageCollector::Host` interface or type to make this more self-documenting.  Can still pass in an inline definition of it just as you are now (which is nice).

src/dag/write-impl.ts
```diff
@@ -85,61 +86,25 @@ export class Write extends Read {
   }
 
   async commit(): Promise<void> {
-    await this.collectGarbage();
+    const garbargeCollector = new GarbargeCollector(
+      this.setRefCount.bind(this),
```
arv, line 90: This seems like it could be done better. Maybe inline?

src/dag/lazy-store.ts
```diff
@@ -0,0 +1,350 @@
...
+import type {ReadonlyJSONValue} from '../mod';
+import {GarbargeCollector, HeadChange} from './gc';
+
+export class LazyStore implements Store {
+  private readonly _rwLock = new RWLock();
+  private readonly _heads = new Map<string, Hash>();
+  private readonly _tempChunks = new Map<Hash, Chunk>();
+  private readonly _sourceChunksCache: ChunksCache;
+  private readonly _sourceStore: Store;
+  private readonly _refCounts = new Map<Hash, number>();
```
aboodman, line 37: Did you consider having `LazyStore` contain a memstore and thereby share more of the GC code btwn two implementations? Seems like you wouldn't need so many callbacks. What are the downsides of that approach?

arv (commented):
I like this approach.

I wonder if there is a solution that uses JS GC instead of us refcounting the lazy store... thinking

src/dag/lazy-store.ts
arv: but where does stuff come from? the RWLock callaback have no params.

src/dag/lazy-store.ts
arv: I think that would be simpler too but it will be less efficient. But lets care about performance later when we know what is slow or not.

src/dag/lazy-store.ts
aboodman: I'm the big picture guy @arv 😂.

I dunno, I was thinking maybe  new lock class that takes the thingy to lock as a parameter. Like:

```
const lock = new FancyLock({foo: "bar"});
lock.withLock(async stuff => {
  // do stuff with stuff
});
```

src/dag/gc.ts
arv: I would not do that, objects are heap allocated and thus have some overhead when this is all internal. TS will complain if things are in wrong order and VSCode can show you parameter hints.

src/dag/lazy-store.ts
arv: We can do that but that is a higher abstraction.

src/dag/lazy-store.ts
arv: Or use a delegate... which could be `this`.

src/dag/gc.ts
grgbkr: im thinking an interface that the WriteImpl can just implement.

src/dag/lazy-store.ts
grgbkr: I explored:
1. having a source dag and a mem dag (for the chunk cache/heads/temp) 
2. using a mem kv store for the cache


Neither really fit because:
1. for both kv store and mem dag the locking is wrong for our needs.  we want to put things into the cache on reads, but we don't want reads to grab a write lock.
2. neither works well for doing lru.  mem dag lacks a delete interface.  neither keep track of access order so we would need to do that on top.  
3. the ref counting semantics are slightly different (due to lazy caching and eviction).  look at how refCounts is manipulated in lazystore ChunkCache.  also retrieving of refs during gc has to know not to go to the source store (otherwise it will load things into cache just for ref counts)

As factored, we get to reuse most of the GC logic.

src/dag/lazy-store.ts
grgbkr: done

src/dag/lazy-store.ts
grgbkr: added a comment

```
  /**
   * This lock is used to ensure correct isolation of Reads and Writes.  Multiple Reads
   * are allowed in parallel but only a single Write.  Reads and Writes see a isolated
   * view of the store (corresponding to the Serializable level of transaction isolation
   * defined in the SQL standard).
   *
   * To ensure these semantics the read lock must be acquired when a Read is created
   * and held til it is closed, and a Write lock must be acquired when a Write is
   * created and held til it is commited.
   *
   * Code must have a read or write lock to
   * - read `heads`
   * - read `tempChunks`
   * - read `_sourceStore`
   * - read and write `_sourceChunksCache`
   * - read and write `_refCounts`
   * and must have a write lock to
   * - write `heads`
   * - write `tempChunks`
   */
```

PR #766, opened 2021-12-16
refactor: Extract interface for dag.Store, dag.Read, and dag.Write

### LINKED ISSUE #671: Implement Dueling Dags

grgbkr:
In preparation for adding a dag.LazyStore implementation, extract an interface for the dag Store.


src/db/read.ts
```diff
@@ -153,10 +153,9 @@ export async function readCommit(
   }
 
   const commit = await commitFromHash(hash, dagRead);
-  const map =
-    dagRead instanceof dag.Write
-      ? new BTreeWrite(dagRead, commit.valueHash)
-      : new BTreeRead(dagRead, commit.valueHash);
+  const map = dag.isWrite(dagRead)
```
aboodman, line 156: 1. I'd check that the third item in the tuple is actually used somewhere or that if it is used, it's not an easy cleanup. This `readCommit()` method dates from waaaay back and I remember when I was in here awhile ago feeling like maybe it could go away. Don't remember the details.

2. If not, would it work to add a `btreeRead()` method to `dag.Read` which is overridden for `dag.Write` to actually return a write? Still a little weird to have a method called `...read` return a `write` but we're already in that situation sort of with this function as it is. Perhaps some better naming could also help ... `btreeOperation` or some such?

grgbkr: What do you think of: https://github.com/rocicorp/replicache/pull/767. 

I don't think we want to add btree methods to the dag store, since btree is at a higher abstraction layer.

arv:
LGTM

PR #769, opened 2021-12-16
refactor: Extract ref count updating logic out of dag WriteImpl so it can be reused by lazy dag store.

grgbkr:
Pulls Garbage Collection logic into its own module, so that it can be shared with the upcoming lazy dag store implementation for Simplified Dueling Dags.

Interface:
```ts
export type HeadChange = {
  new: Hash | undefined;
  old: Hash | undefined;
};

export type RefCountUpdates = Map<Hash, number>;

export interface GarbageCollectionDelegate {
  getRefCount: (hash: Hash) => Promise<number>;
  getRefs: (hash: Hash) => Promise<readonly Hash[] | undefined>;
}

/**
 * Computes how ref counts should be updated when a dag write is commited.
 * Does not modify the dag store.
 * @param headChanges Heads that were changed by the dag write.
 * @param putChunks Chunks that were put by the dag write.
 * @param delegate Delegate used for loading ref information from the dag store.
 * @returns Map from chunk Hash to new ref count.  Chunks with a new ref count of 0 should
 * be deleted.  All hashes in `putChunks` will have an entry (which will be zero if a
 * newly put chunk is not reachable from any head).
 */
export async function computeRefCountUpdates(
  headChanges: Iterable<HeadChange>,
  putChunks: ReadonlySet<Hash>,
  delegate: GarbageCollectionDelegate,
): Promise<RefCountUpdates> 
```

Part of #671


src/dag/gc.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  const refCountUpdates = await computeRefCountUpdates(
+    [{old: hashes['r'], new: hashes['a']}],
+    new Set(),
+    delegate,
+  );
+  expectRefCountUpdates(refCountUpdates, {
+    r: 0,
+    a: 1,
+    b: 0,
+    c: 1,
```
arv, line 157: why isn't `d` in here?

src/dag/gc.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    new Set(Object.values(hashes)),
+    delegate,
+  );
+  expectRefCountUpdates(refCountUpdates, {
+    r: 1,
+    a: 1,
+    b: 1,
+    c: 0,
+    d: 0,
+  });
```
arv, line 83: I think this test would be easier to read if it tested the current ref count value and not only the ref count updates

arv:
LGTM

PR #770, opened 2021-12-17
refactor: move getSizeOfValue from src/btree/ to src/json.ts

grgbkr:
### Problem
`getSizeOfValue` is needed by the upcoming Lazy DagStore for Simplified Dueling Dags.  It is needed for implementing LRU caching with a size limit.  However the dag/ directory should not depend on the btree/ directory, as dag is at a lower abstraction layer than btree.

### Solution
Move `getSizeOfValue` to src/json.ts.  This is a logic place of the function as it computes the size of a `ReadonlyJsonValue`.


arv (commented):
LGTM

PR #778, opened 2022-01-12
refactor: Delete unused slurp and merge kv.MemStore into kv.TestMemStore

grgbkr:
With the integration of dag.LazyStore slurp is no longer used, and kv.MemStore is only 
used via kv.TestMemStore in tests.


src/kv/test-mem-store.ts
```diff
@@ -1,8 +1,50 @@
...
-import {MemStore} from './mem-store';
-import type {Value} from './store';
+import {RWLock} from '../rw-lock';
+import type {Read, Store, Value, Write} from './store';
+import {deleteSentinel, WriteImplBase} from './write-impl-base';
 
-export class TestMemStore extends MemStore {
+export class TestMemStore implements Store {
+  // protected to allow test sub class to use it.
+  protected readonly _map: Map<string, Value> = new Map();
```
arv: Seems like this can be private now?

arv (approved):

Issue #595, opened 2021-09-29
R3 Performance Envelope


aboodman:
A [*Performance Envelope*](https://en.wikipedia.org/wiki/Flight_envelope) is the range of conditions under which an aircraft is designed and engineered to work properly. Pilots need to know the envelope of the aircraft they are flying in order to use it properly and for the right applications.

A perf envelope is super useful for a piece of software like Replicache too, as it lets application developers understand how they should (and should not) use the product.

This bug specifies our current target perf envelope and our progress toward achieving it:

data sizes
- Work very well with 100MB of data, acceptably with 1GB. Larger datasets out of scope.

write/sub/read
- 95% of writes/sub/read cycle complete in <1ms with 100 subscriptions, 5 dirty that read 10KB each.

startup
- @100mb: 95% Read first 100kb < 100ms
- @1gb: 95% Read first 100kb in < 1000ms

sync
- TODO: What should our target be for how long it takes to apply a 100MB initial sync? 1GB?
- TODO: What should our target be for how long it takes to apply a small incremental sync
- Propagate mutations between clients with low latency when online. This of course depends on backend, and topology, but collaborating users on Replicache should typically see each others’ changes in <1s, 2s max

offline
- Propagate mutations between tabs even when offline. It’s common for users to accidentally have multiple tabs with same app/doc open. Changes in one tab should show up in other and not be lost. It’s fine to do this on a slow timer like 5s or just on focus/tabshow.
- All pending mutations applied must be pushed to server when back online. For example, if pending mutations are applied in a tab while offline and that tab is closed (or crashes) while offline, there must be a mechanism that ensures its mutations are pushed when back online.

aboodman:
@phritz can you please add targets for the two sync items?

aboodman:
> @1gb: 95% Read first 100kb in < 1000ms

A customer points out that this is never going to make sense for users. Never going to want 1gb in Replicache if it has this effect on startup.

I don't think we actually expect startup to be influenced by cache size, so should we just say:

> @1gb: 95% Read first 100kb in < 100ms

(e.g., we expect startup to be flat past 100MB)

aboodman:
@phritz

PR #780, opened 2022-01-13
feat: Add benchmarks for startup from persisted state 

grgbkr:
Add two benchmarks for startup from persisted state based on the Replicache performance envelope (#595).

1. Init replicache and read 100 KB of data from 100 MB of persisted state using `get`s of random keys.
2. Init replicache and read 100 KB of data from 100 MB of persisted state using `scan` starting at a random key.


perf/replicache.ts
```diff
@@ -48,6 +50,151 @@ export function benchmarkPopulate(opts: {
...
+        },
+        httpRequestInfo: {
+          httpStatusCode: 200,
+          errorMessage: '',
+        },
+      };
+    },
+  });
+
+  let initialPullComplete = false;
```
arv: You could have used subscribe here instead but it does not really matter.

perf/replicache.ts
```diff
@@ -48,6 +50,151 @@ export function benchmarkPopulate(opts: {
...
+    );
+  }
+  await repForStore.persist();
+  await repForStore.close();
+}
+
+export function benchmarkStartupUsingBasicReadsFromPersistedData(opts: {
+  numKeysPersisted: number;
+  numKeysToRead: number;
+  useMemstore: boolean;
```
arv: useMemstore is only used in the name. Consider dropping it

perf/replicache.ts
```diff
@@ -48,6 +50,151 @@ export function benchmarkPopulate(opts: {
...
+      console.log(getCount);
+      await rep.close();
+    },
+  };
+}
+
+export function benchmarkStartupUsingScanFromPersistedData(opts: {
+  numKeysPersisted: number;
+  numKeysToRead: number;
+  useMemstore: boolean;
```
arv: ditto

src/dag/lazy-store.ts
```diff
@@ -474,6 +478,28 @@ class ChunksCache {
       if (this._size <= this._cacheSizeLimit) {
         break;
       }
+      const cacheRefs = new Set<Hash>();
```
arv: What is this change about?

src/kv/idb-store.ts
```diff
@@ -164,6 +164,7 @@ function objectStore(tx: IDBTransaction): IDBObjectStore {
 }
 
 async function openDatabase(name: string): Promise<IDBDatabase> {
+  console.log('opening ' + name);
```
arv: remove

arv (changes requested):

PR #781, opened 2022-01-14
Adds a timestamp field to mutations.

aboodman:
Note: Format change. Not changing REPLICACHE_FORMAT_VERSION constant
because we should only do so once per release by policy, so that
should be something we do during release.


src/db/visitor.test.ts
```diff
@@ -83,6 +83,7 @@ test('test that we get to the data nodes', async () => {
       fakeHash('none'),
       prevCommit.valueHash,
       prevCommit.indexes,
+      88,
```
arv, line 86: Please do not use this number for tests... it is associated with a symbol of hate.

src/replicache.ts
```diff
@@ -225,6 +227,8 @@ export class Replicache<MD extends MutatorDefs = {}> {
    */
   pusher: Pusher;
 
+  now: Now = () => performance.now();
```
arv, line 230: New public API? Why is this exposed?

src/replicache.ts
```diff
@@ -1156,6 +1163,7 @@ export class Replicache<MD extends MutatorDefs = {}> {
           name,
           mutatorImpl,
           args,
+          performance.now(),
```
arv, line 1166: should this be `this.now()`?

arv (commented):
Needs more work

src/db/visitor.test.ts
aboodman: Whoops, forgot about that. For me it always means the number of miles per hour the delorean has to go to activate the flux capacitor. Will avoid in the future.

src/replicache.ts
aboodman: Mistake.

src/replicache.ts
aboodman: No, I thought at first that `Now` would have to be injected for testing, but you are already controlling the timers in the tests with monkey patching.

aboodman:
Addressed in https://github.com/rocicorp/replicache/pull/783.

PR #792, opened 2022-01-24
feat: Simplified Dueling Dags - Mutation Recover - Add mutationID and  lastServerAckdMutationID to Client.

grgbkr:
Add `mutationID` and  `lastServerAckdMutationID` to `Client` and update `initClient` and `persist` to
write them appropriately.  These new fields will be used by other clients to determine if a client has 
pending mutations (persisted local mutations unacknowledged by the server) that it can push on the 
other client's behalf.  We will refer to this process as “mutation recovery”, as one client is recovering the
 mutations of another client, by reading them from the other client’s perdag stage and pushing on the 
other client’s behalf. 

See [Mutation Recovery design](https://www.notion.so/Mutation-Recovery-Avoiding-Mutation-Loss-using-PerDag-state-f54025b52cbc435692abca3307947d15). 

Part of #671


src/mod.ts
```diff
@@ -1,6 +1,7 @@
 export {Replicache} from './replicache';
 export {TransactionClosedError} from './transaction-closed-error';
 
+export type {Hash} from './hash';
```
arv: No.

We should not expose this in the public API

src/persist/persist.ts
```diff
@@ -41,21 +42,42 @@ export async function persist(
...
 }
 
 async function gatherTempChunks(
   memdag: dag.Store,
-): Promise<[ReadonlyMap<Hash, dag.Chunk>, Hash]> {
+): Promise<
+  [
+    ReadonlyMap<Hash, dag.Chunk>,
+    Hash,
+    {mutationID: number; lastMutationID: number},
```
arv: Either make this:

```ts
type _ = [
    map: ReadonlyMap<Hash, dag.Chunk>,
    hash: Hash,
    mutationID: number,
    lastMutationID: number,
]
```

or

```ts
type _ = {
    map: ReadonlyMap<Hash, dag.Chunk>,
    hash: Hash,
    mutationID: number,
    lastMutationID: number,
}
```

mixing object and array seems a bit ugly

arv: the benefit of tuple over object is that it compresses better. the benefit of object is that it reads better...

src/persist/persist.ts
```diff
@@ -96,13 +118,16 @@ async function writeFixedChunks(
   fixedChunks: FixedChunks,
   mainHeadHash: Hash,
   clientID: string,
+  mutationIDs: {mutationID: number; lastMutationID: number},
```
arv: same here... just make these into two parameters?

arv (approved):
But please undo the changes to mod.ts

We should probably add a comment in there or rename the file. The name was picked because mod.ts has become the defacto standard name for the public API in deno.

src/mod.ts
grgbkr: I see.  So just src/mod.ts is the public API, correct?  The other mod.ts files are for cross directory export/import?

src/persist/persist.ts
grgbkr: done

src/persist/persist.ts
grgbkr: done

src/mod.ts
arv: That is correct

arv (approved):
LGTM

arv (approved):

PR #797, opened 2022-01-27
refactor: Extract reauth retry logic from push and pull into shared method.

grgbkr:
Extract reauth retry logic from push and pull into shared method.  This reduces duplication and also will allow for reuse of this logic for mutation recovery.

### DIFF
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -76,7 +76,7 @@ export function makeIdbName(name: string, schemaVersion?: string): string {
  * The maximum number of time to call out to getAuth before giving up
  * and throwing an error.
  */
-export const MAX_REAUTH_TRIES = 8;
+const MAX_REAUTH_TRIES = 8;
 
 const PERSIST_TIMEOUT = 1000;
 
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -42,6 +42,7 @@ import {
 import {Lock} from './rw-lock';
 import * as persist from './persist/mod';
 import {requestIdle} from './request-idle';
+import type {HTTPRequestInfo} from './http-request-info';
 
 export type BeginPullResult = {
   requestID: string;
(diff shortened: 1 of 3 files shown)


src/replicache.ts
```diff
@@ -72,7 +73,7 @@ export function makeIdbName(name: string, schemaVersion?: string): string {
 }
 
 /**
- * The maximum number of time to call out to getDataLayerAuth before giving up
+ * The maximum number of time to call out to getAuth before giving up
  * and throwing an error.
  */
 export const MAX_REAUTH_TRIES = 8;
```
arv: Maybe no need to export now?

arv (approved):
LGTM

src/replicache.ts
```diff
@@ -785,69 +763,110 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
-              this.getPushAuth!());
-          if (auth !== null && auth !== undefined) {
-            this.auth = auth;
-            // Try again now instead of waiting for next push.
-            return await this._invokePush(maxAuthTries - 1);
-          }
-        }
-
-        return httpRequestInfo.httpStatusCode === 200;
+      if (auth === null || auth === undefined) {
```
grgbkr, line 819: @arv this matches the existing logic, but its not clear to me why it is correct.  If getAuth returns null or undefined, why do we give up rather than retrying call getAuth up to MAX_REAUTH_TRIES?

src/replicache.ts
grgbkr: good idea, done.

src/replicache.ts
arv: The idea was that `getAuth()` would try to authenticate and return the new auth token. When we get a new auth token we try again. Returning null/undefined from `getAuth()` is a way to signal that we have given up. If we instead tried again the app might show another login prompt.

PR #798, opened 2022-01-27
feat: Simplified Dueling Dags - Mutation Recovery - Add an optional parameter to beingPull for disabling the creation of a sync branch from the pull response

grgbkr:
For Mutation Recover we need to be able to pull to confirm mutations have been applied on the server by 
looking at the responses `lastMutationID`, but we do not want to apply the response to the DAG.  Add an 
option to beginPull to not create a sync branch from the pull response.

Also add the `PullResponse` to `BeginPullResponse`, as it will be need by Mutation Recovery to get the 
`lastMutationID`. 

Part of #671

### DIFF
--- a/src/sync/pull.test.ts
+++ b/src/sync/pull.test.ts
@@ -89,6 +89,7 @@ test('begin try pull', async () => {
 
   type Case = {
     name: string;
+    createSyncBranch?: boolean;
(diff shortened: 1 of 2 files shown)


arv (commented):
LGTM

PR #799, opened 2022-01-27
feat: Simplified Dueling Dag - Mutation Recovery - Implement the mutation recovery process.

grgbkr:
### Problem 
With Simplified Dueling Dags mutations that have not been synced to the server when a tab is 
unloaded (or frozen and never unfrozen) are lost.  This can occur in common user flows, and 
will result in unexpected data loss.  The impact is worst when the user has been offline or has 
a flakey connection as there will be more local mutations that have not been synced.   Cases 
where this will occur:

- Refresh before changes have been pushed
- Close before changes have been pushed
- Navigate away before changes have been pushed
- Tab backgrounded and frozen before changes have been pushed (seems unlikely) and tab is not revisited before client is gc’d
- Tab crash before changes have been pushed

### Solution
Replicache clients will try to recover mutations from other Replicache client's perdag state.   
A Replicache client can recover another Replicache client's mutations if the other clients has 
the same name (and thus can share auth), the same domain, and a Replicache format and 
schema version understood by the client.   A Replicache client will try to recover other
clients' mutation at startup, reconnection and on a 5 minute interval.

See full design at https://www.notion.so/replicache/Mutation-Recovery-Avoiding-Mutation-Loss-using-PerDag-state-f54025b52cbc435692abca3307947d15

### DIFF
--- a/src/replicache-mutation-recovery.test.ts
+++ b/src/replicache-mutation-recovery.test.ts
@@ -20,6 +20,7 @@ import {uuid} from './sync/uuid';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-expect-error
 import fetchMock from 'fetch-mock/esm/client';
+import {assertJSONObject, JSONObject} from './json';
 
 initReplicacheTesting();
 
@@ -122,18 +123,11 @@ async function testRecoveringMutationsOfClient(schemaVersions: {
 
   fetchMock.reset();
   fetchMock.post(pushURL, 'ok');
-  fetchMock.post(
-    {
-      url: pullURL,
-      body: {clientID: client1ID},
-      matchPartialBody: true,
-    },
-    {
-      cookie: 'cookie1',
-      lastMutationID: client1.mutationID,
-      patch: [],
-    },
-  );
+  fetchMock.post(pullURL, {
+    cookie: 'pull_cookie_1',
+    lastMutationID: client1.mutationID,
+    patch: [],
+  });
 
   await rep.recoverMutations();
 
@@ -157,6 +151,16 @@ async function testRecoveringMutationsOfClient(schemaVersions: {
     ],
     pushVersion: 0,
     schemaVersion: schemaVersionOfClientWPendingMutations,
+  });
+
+  const pullCalls = fetchMock.calls(pullURL);
+  expect(pullCalls.length).to.equal(1);
+  expect(await pullCalls[0].request.json()).to.deep.equal({
+    clientID: client1ID,
+    schemaVersion: schemaVersionOfClientWPendingMutations,
+    cookie: 'cookie_1',
+    lastMutationID: client1.lastServerAckdMutationID,
+    pullVersion: 0,
   });
 
   const updatedClient1 = await testPerdag.withRead(read =>
@@ -256,49 +260,43 @@ test('successfully recovering mutations of multiple clients with mix of schema v
   );
   assertNotUndefined(client4);
 
+  const pullRequestJsonBodies: JSONObject[] = [];
   fetchMock.reset();
   fetchMock.post(pushURL, 'ok');
   fetchMock.post(
-    {
-      url: pullURL,
-      body: {clientID: client1ID},
-      matchPartialBody: true,
-      name: 'pullClient1',
-    },
-    {
-      cookie: 'cookie1',
-      lastMutationID: client1.mutationID,
-      patch: [],
-    },
-  );
-  fetchMock.post(
-    {
-      url: pullURL,
-      body: {clientID: client3ID},
-      matchPartialBody: true,
-      name: 'pullClient3',
-    },
-    {
-      cookie: 'cookie3',
-      lastMutationID: client3.mutationID,
-      patch: [],
-    },
-  );
-  fetchMock.post(
-    {
-      url: pullURL,
-      body: {clientID: client4ID},
-      matchPartialBody: true,
-      name: 'pullClient4',
-    },
-    {
-      cookie: 'cookie4',
-      lastMutationID: client4.mutationID,
-      patch: [],
+    pullURL,
+    async (url: string, options: RequestInit, request: Request) => {
+      const requestJson = await request.json();
+      assertJSONObject(requestJson);
+      pullRequestJsonBodies.push(requestJson);
+      const {clientID} = requestJson;
+      switch (clientID) {
+        case client1ID:
+          return {
+            cookie: 'pull_cookie_1',
+            lastMutationID: client1.mutationID,
+            patch: [],
+          };
+        case client3ID:
+          return {
+            cookie: 'pull_cookie_2',
+            lastMutationID: client3.mutationID,
+            patch: [],
+          };
+        case client4ID:
+          return {
+            cookie: 'pull_cookie_3',
+            lastMutationID: client4.mutationID,
+            patch: [],
+          };
+        default:
+          throw new Error(`Unexpected pull ${requestJson}`);
+      }
     },
   );
 
   await rep.recoverMutations();
+
   const pushCalls = fetchMock.calls(pushURL);
   expect(pushCalls.length).to.equal(3);
   expect(await pushCalls[0].request.json()).to.deep.equal({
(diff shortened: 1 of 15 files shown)


src/replicache-options.ts
```diff
@@ -53,6 +53,13 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
    */
   schemaVersion?: string;
 
+  /**
+   * Previous schema versions of this application's data. This enables recovery
+   * of pending mutations from and garbage collection of data persisted
+   * with one of these schema versions.
+   */
+  previousSchemaVersions?: string[];
```
arv: Can we not discover this automatically. Having to provide old schema version is annoying.

src/replicache.ts
```diff
@@ -55,10 +56,26 @@ export type Poke = {
   pullResponse: PullResponse;
 };
 
+enum MutationRecoveryStatus {
```
arv: `const enum` or no enum.

src/replicache.ts
```diff
@@ -176,6 +193,7 @@ export class Replicache<MD extends MutatorDefs = {}> {
 
   /** The schema version of the data understood by this application. */
   schemaVersion: string;
+  previousSchemaVersions: string[];
```
arv: Why public read/write property? All public API should have JSDoc

Also, we should make schemaVersion readonly. Changing it at runtime does not seem meaningful

src/replicache.ts
```diff
@@ -1237,6 +1258,201 @@ export class Replicache<MD extends MutatorDefs = {}> {
       return {result, ref};
     });
   }
+
+  private async _recoverMutations(): Promise<void> {
+    if (this._recoveringMutations) {
```
arv: TS is not able to reason about this async code. You need a `// @ts-ignore` here

src/replicache.ts
```diff
@@ -1237,6 +1258,201 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+    try {
+      this._recoveringMutations = true;
+      const ok = await this._recoverMutationsFromPerdag(
+        this._perdag,
+        this.schemaVersion,
+      );
+      if (ok) {
+        for (const schemaVersion of this.previousSchemaVersions) {
+          const perKvStore = new IDBStore(
+            makeIdbName(this.name, this.schemaVersion),
```
arv: TODO: Also check old REPLICACHE_FORMAT_VERSION?

src/replicache.ts
```diff
@@ -1237,6 +1258,201 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+            perKvStore,
+            dag.throwChunkHasher,
+            assertNotTempHash,
+          );
+          const ok = await this._recoverMutationsFromPerdag(
+            perdag,
+            schemaVersion,
+          );
+          if (!ok) {
+            break;
```
arv: Not sure about this. Why should it exit the loop here?

src/replicache.ts
```diff
@@ -1237,6 +1258,201 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+    }
+  }
+
+  private async _recoverMutationsFromPerdag(
+    perdag: dag.Store,
+    schemaVersion: string,
+  ): Promise<boolean> {
+    let clientMap: persist.ClientMap = await perdag.withRead(async read => {
+      return await persist.getClients(read);
+    });
```
arv:
```suggestion
    let clientMap: persist.ClientMap = await perdag.withRead(read => 
      persist.getClients(read));
```

src/replicache.ts
```diff
@@ -1237,6 +1258,201 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+        if (!clientIdsVisited.has(clientID)) {
+          clientIdsVisited.add(clientID);
+          const result = await this._recoverMutationsOfClient(
+            client,
+            clientID,
+            perdag,
+            schemaVersion,
+          );
+          if (result.status === MutationRecoveryStatus.SUCCESS) {
+            newClientMap = result.newClientMap;
```
arv: or

```suggestion
            ({newClientMap} = result);
```

arv (commented):
Looking pretty good

src/replicache.ts
```diff
@@ -1237,6 +1258,201 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+    return !connectivityError;
+  }
+
+  private async _recoverMutationsOfClient(
+    client: persist.Client,
+    clientID: sync.ClientID,
+    perdag: dag.Store,
+    schemaVersion: string,
+  ): Promise<MutationRecoveryResult> {
+    await this._ready;
```
grgbkr: this should be moved (or copied to _recoverMutations

src/replicache-options.ts
grgbkr: yeah we just need to store it in either local storage or indexeddb due to lack of support for discovering these on firefox.

Probably indexedb is the way to go as the lack of concurrency control in localstorage makes it tricky to avoid clobbering when trying to add to a map or list.

src/replicache.ts
grgbkr: done

src/replicache.ts
grgbkr: yeah the perviousSchemaVersions is not fully worked out. I think I'm going to take your advice and do this internally so it doesnt have to be part of the public api.

src/replicache.ts
grgbkr: the issue is I haven't added so code to call `_recoverMutations` yet.

src/replicache.ts
grgbkr: done

src/replicache.ts
grgbkr: done

src/replicache.ts
grgbkr: done

src/replicache.ts
grgbkr: The idea is to fail fast on the mutation recovery process if we hit a connectivity error.  _recoverMutationsFromPerdag returns false only when hitting a connectivity error.  When we transition from offline -> online we will restart the process by calling this._recoverMutations.

I consider it a connectivity error if push or pull got an error and this.online is now false.  For other errors like the server returning a 500, we do not fail fast.

It may be more reliable to not fail fast.  If for some reason recovering a particular client consistently causes an error that we incorrectly interpret as a connectivity error, we will never recover clients after it in the ClientMap.  The tradeoff is we do some perdag reading and pushes that we expect to fail because we are "offline".

src/replicache.ts
```diff
@@ -55,10 +56,26 @@ export type Poke = {
...
+const enum MutationRecoveryStatus {
+  NO_MUTATIONS_TO_RECOVER,
+  SUCCESS,
+  ERROR,
+}
+
+type MutationRecoveryResult = {
+  status: MutationRecoveryStatus;
+  /**
+   * When `status` is `SUCCESS`, the new `ClientMap` state, otherwise
```
arv: can you use a discriminated union here?

grgbkr: decided to simplify and not fail fast on connectivity issues, this led to not needing this more complicated return structure.

src/replicache.ts
grgbkr: done

src/replicache.ts
grgbkr: decided to simplify and not fail fast on connectivity issues, this is simpler and more robust.

src/replicache-mutation-recovery.test.ts
```diff
@@ -0,0 +1,404 @@
...
+initReplicacheTesting();
+
+const dagsToClose: dag.Store[] = [];
+const dbsToDelete: Set<string> = new Set();
+
+teardown(async () => {
+  for (const dagToClose of dagsToClose) {
+    try {
+      await dagToClose.close();
+      // eslint-disable-next-line no-empty
```
arv: or `console.error(e)`?

src/replicache-mutation-recovery.test.ts
```diff
@@ -0,0 +1,394 @@
...
+  );
+  assertNotUndefined(client1);
+
+  fetchMock.reset();
+  fetchMock.post(pushURL, 'ok');
+  fetchMock.post(
+    {
+      url: pullURL,
+      body: {clientID: client1ID},
+      matchPartialBody: true,
```
grgbkr: example

arv (commented):
LGTM

PR #801, opened 2022-01-28
chore: Delete change broadcast code

grgbkr:
In the new unified storage model of simplified dueling dags changes are shared between 
tabs via syncing with the server.   This code is no longer needed, possibly impacting
performance and causing spurious firing of subscriptions.


aboodman:
yay. lgtm.

arv (approved):
Thanks

PR #802, opened 2022-01-28
feat: Track Replicache IndexedDB databases in another IndexedDB database for mutation recovery and db gc.

grgbkr:
### Problem
We need to be able to find old Replicache IndexedDB databases (i.e. databases with previous schema 
versions or replicache format versions), so that we can recover mutations from them and also GC them.

### Solution
Keep track of Replicache IndexedDB databases in a IndexedDB database. 

Unfortunately Firefox does not implement [IDBFactory.databases](https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/databases), or we would use that api.

IndexedDB is used over LocalStorage because LocalStorage's lack of concurrency control makes
it very difficult to avoid write clobbering when updating a list or map.

### DIFF
--- a/src/persist/idb-databases-store.ts
+++ b/src/persist/idb-databases-store.ts
@@ -13,6 +13,8 @@ export type IndexedDBDatabase = {
   schemaVersion: string;
 };
 
+export type IndexedDBDatabaseRecord = Record<IndexDBName, IndexedDBDatabase>;
+
 function assertIndexedDBDatabaseRecord(
   value: unknown,
 ): asserts value is IndexedDBDatabaseRecord {
@@ -31,8 +33,6 @@ function assertIndexedDBDatabase(
   assertNumber(value.replicacheFormatVersion);
   assertString(value.schemaVersion);
 }
-
-export type IndexedDBDatabaseRecord = Record<IndexDBName, IndexedDBDatabase>;
 
 export class IDBDatabasesStore {
   private readonly _kvStore: kv.Store;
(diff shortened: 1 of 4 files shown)


src/persist/idb-databases-store.ts
```diff
@@ -0,0 +1,70 @@
...
+      await write.put(KEY, dbRecord);
+      await write.commit();
+      return dbRecord;
+    });
+  }
+
+  getDatabases(): Promise<IndexedDBDatabaseRecord> {
+    return this._kvStore.withRead(async read => {
+      return this._getDatabases(read);
+    });
```
arv:
```suggestion
    return this._kvStore.withRead(read => this._getDatabases(read));
```

src/replicache.ts
```diff
@@ -291,6 +292,7 @@ export class Replicache<MD extends MutatorDefs = {}> {
       this._memdagHashFunction(),
       assertHash,
     );
+    this._idbDatabases = new persist.IDBDatabasesStore();
```
arv: could move this to line 212

arv (approved):

arv:
I really like how you are using the kv.Store to handle this. Beautiful!

src/persist/idb-databases-store.ts
grgbkr: wish prettier would catch these.  done.

src/replicache.ts
grgbkr: done

PR #808, opened 2022-02-04
fix: fix benchmarks broken by mutation recovery change

grgbkr:
These were broken by 1538c424e2369c36c5a41c7876f0dad338e59768. 

three fixes.
1. close other indexeddbs we try to recover
2. delete IDBDatabasesStore indexeddb db after each benchmark.  this way mutationRecovery is not try to recover a ton of dbs during benchmarks. 
3. make the indexeddb deletion code in perf tests more robost.  There is no way to wait for an indexeddb close to complete (the api is silently async).  If you call delete when a close is in process, it fails with a onblocked event.  Add code that will delay 100ms and then retry delete after a onblocked event (up to 10 retries).

### DIFF
--- a/perf/perf.ts
+++ b/perf/perf.ts
@@ -189,7 +189,7 @@ export async function runBenchmarkByNameAndGroup(
           : formatAsBenchmarkJS(result),
     };
   } catch (e) {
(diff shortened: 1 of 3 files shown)


arv (commented):
LGTM

PR #810, opened 2022-02-04
fix: Simplified Dueling Dags - Mutation Recovery - Do not recover mutation from clients with a different Replicache name

grgbkr:
Problem
======
A client _**MUST NOT**_ recover mutations from a client with a different Replicache name.  This is because a client uses its auth to push the mutations.  This is safe for client's with the same name as they are for the same user.  However, pushing on behalf of a client with a different name is very bad, as it will apply the mutations for a different user.

Solution
======
Add Replicache name to the IndexedDBDatabase records, and only recover mutations for clients with the same Repliache name.  Add a test for this behavior.

Also adds versioning to IDBDatabasesStore for easing handling of future format changes of  IndexedDBDatabase records.


arv (commented):
LGTM

PR #813, opened 2022-02-06
add licensing client

phritz:
Doesn't do much for now, this is just getting the integration done. If developing locally you can run `npm link` to link the get-license command and then `npx get-license` to fetch a new, unpersisted, useless license key. But hey, it works end to end!

### DIFF
--- a/tool/get-license.ts
+++ b/tool/get-license.ts
@@ -1,0 +1,12 @@
+#!/usr/bin/env node
+
+// If developing Replicache locally, run `npm link` to link the
+// command. Then you can run this command with `npx get-license`.
+// You should just be able to run `npx get-license` if you are
+// in a package that has Replicache as a local depenency. (Note:
+// if not, the binary lives under node_modules/.bin/, we would
+// need to run it from there instead.)
+
+import * as licensingCLI from '@phritz/licensing/src/cli';
+
+licensingCLI.main();
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -21,6 +21,11 @@
     "lib": ["dom", "esnext"],
     "preserveSymlinks": true
   },
-  "include": ["src/**/*.ts", "perf/**/*.ts", "perf/**/*.js"],
+  "include": [
+    "src/**/*.ts",
+    "perf/**/*.ts",
+    "perf/**/*.js",
+    "node_modules/@phritz/licensing/**/*.ts"
+  ],
   "exclude": ["node_modules", "dist/**/*", "build/**/*"]
 }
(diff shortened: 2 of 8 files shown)


tool/get-license.ts
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env node
+
+// If developing Replicache locally, run `npm link` to link the
+// command. Then you can run this command with `npx get-license`.
+// You should just be able to run `npx get-license` if you are
+// in a package that has Replicache as a local depenency. (Note:
+// if not, the binary lives under node_modules/.bin/, we would
+// need to run it from there instead.)
+
+import * as licensingCLI from '@phritz/licensing/src/cli';
```
arv, line 10: The export should be `'@phritz/licensing/cli';`  (no `src`)

arv: You can achieve this wit h the `exports` field in the package.json for `@phritz/licensing`

tsconfig.json
```diff
@@ -21,6 +21,11 @@
     "lib": ["dom", "esnext"],
     "preserveSymlinks": true
   },
-  "include": ["src/**/*.ts", "perf/**/*.ts", "perf/**/*.js"],
+  "include": [
+    "src/**/*.ts",
+    "perf/**/*.ts",
+    "perf/**/*.js",
+    "node_modules/@phritz/licensing/**/*.ts"
```
arv, line 28: This is not the usual way to do this.

The `@phritz/licensing` package includes the js and .d.ts files so TS knows how to include that.

...but I guess it is a private package so we could even use a github dependency in our package.json. Whatever is easiest.

tool/get-license.ts
phritz: Per discussion with @aboodman we are publishing the licensing package as typescript sources and depending on its sources as a dev dependency in replicache. We are not depending on its built output in replicache so I think I the I need the `src` here? The compiler complains about not finding it if I omit the src.

tsconfig.json
phritz: Apparently I can remove it, so I just did in the new PR.

tool/get-license.ts
arv: That sounds right to me.

I think, to make that even simpler we can use a github dependency in package.json

phritz: I'm going to leave this as an exercise for the future. (I would, however, be interested in a very easy way to have a common package that we could put things like .eslint, prettierrc, asserts.ts, etc in and share. Git submodules?)

PR #820, opened 2022-02-07
feat: Simplified Dueling Dags - Mutation Recovery - Make recovery robust to errors and exit early on close.

grgbkr:
**Problem**
If the Mutation Recovery process stops on the first error encountered.  This means a single problematic db or client can prevent recovery of all other clients.  

**Solution**
Updates Mutation Recovery logic to be more robust against errors.  If an error occurs recovering a particular 
client or db, the logic will now log that error, and continue trying to recover other clients/dbs. 

Adding the above robustness requires the process to handle the Replicache instance being closed more explicitly.  Previously the process would stop on the first error encountered due to the Replicache intance being closed. 
This change updates the logic to check if this Replicache instance is closed before processing each db, and 
each client inside each db, and exits early if this Replicache instance is closed.

### DIFF
--- a/src/replicache-mutation-recovery.test.ts
+++ b/src/replicache-mutation-recovery.test.ts
@@ -11,6 +11,7 @@ import type * as db from './db/mod';
 import * as dag from './dag/mod';
 import * as persist from './persist/mod';
 import * as kv from './kv/mod';
+import type * as sync from './sync/mod';
 import {assertHash, assertNotTempHash, makeNewTempHashFunction} from './hash';
 import {assertNotUndefined} from './asserts';
 import {expect} from '@esm-bundle/chai';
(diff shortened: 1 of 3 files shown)


src/replicache-mutation-recovery.test.ts
```diff
@@ -86,6 +88,24 @@ async function createAndPersistClientWithPendingLocal(
   return localMetas;
 }
 
+function createPushBody(
+  client1ID: string,
```
arv, line 92: s/client1ID/clientID/

arv (commented):
lgtm

src/replicache-mutation-recovery.test.ts
grgbkr: Done: https://github.com/rocicorp/replicache/pull/825

thanks for the review

PR #821, opened 2022-02-07
feat: Simplified Dueling Dags - Mutation Recovery - Optimize mutation recovery at startup by reusing client map read by client init.

grgbkr:
**Problem**
Mutation recovery regressed our median startup scan benchmark by ~20% (25 ms to 30 ms).

**Solution**
Try to mitigate by reusing the client map read by `persist.initClient`, rather than reading it in a new IndexedDB transaction.

### DIFF
--- a/src/persist/clients.test.ts
+++ b/src/persist/clients.test.ts
@@ -536,7 +536,15 @@ test('updateClients throws errors if chunk pointed to by clients head does not c
 test('initClient creates new empty snapshot when no existing snapshot to bootstrap from', async () => {
   const dagStore = new dag.TestStore();
   clock.tick(4000);
-  const [clientId, client] = await initClient(dagStore);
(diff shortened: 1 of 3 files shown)


src/persist/clients.test.ts
```diff
@@ -536,7 +536,15 @@ test('updateClients throws errors if chunk pointed to by clients head does not c
 test('initClient creates new empty snapshot when no existing snapshot to bootstrap from', async () => {
   const dagStore = new dag.TestStore();
   clock.tick(4000);
-  const [clientId, client] = await initClient(dagStore);
+  const [clientId, client, clients] = await initClient(dagStore);
+
+  expect(clients).to.deep.equal(
+    new Map(
```
arv, line 542:
```js
new Map([[clientID, client]])
```

arv (commented):
LGTM

PR #822, opened 2022-02-08
fix: Improve isolation of tests' and benchmarks' indexeddb state to reduce flakiness

### LINKED ISSUE #819: Address flakiness of persist/closed database
The following issues happen sporadically.  I think the underlying issue is there is no reliable way to wait for a indexedb db to close, and so if we reuse db names across test the close from test 1 can still be finishing while test 2 has opened the db and tried to start a transaction.

```
src/replicache-persist.test.ts:

 ❌ basic persist & load (failed on Webkit)
      AssertionError: expected undefined to equal 1
        at r (node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:9576:21)
        at node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:251:8
        at p (node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:1410:10)
        at c (node_modules/@esm-bundle/chai/node_modules/.pnpm/chai@4.3.4/node_modules/chai/chai.js:7884:30)
        at src/replicache-persist.test.ts:52:39

src/replicache-subscribe.test.ts:

 🚧 Browser logs on Firefox:
      An error was thrown in a Promise outside a test. Did you forget to await a function or assertion?
      InvalidStateError: IDBDatabase.transaction: Can't start a transaction on a closed database
        at readImpl (src/kv/idb-store.ts:158:16)
        at withRead (src/kv/idb-store.ts:35:26)
        at async*withRead (src/dag/store-impl.ts:36:20)
        at updateClients (src/persist/clients.ts:214:48)
        at writeFixedChunks (src/persist/persist.ts:125:8)
        at persist (src/persist/persist.ts:45:8)
        at async*_persist/< (src/replicache.ts:924:14)
        at run (src/rw-lock.ts:59:17)
        at async*withLock (src/rw-lock.ts:15:11)
        at _persist (src/replicache.ts:923:29)
        at async*_schedulePersist/< (src/replicache.ts:935:17)
        at async*_schedulePersist (src/replicache.ts:936:33)
        at _mutate/< (src/replicache.ts:1175:13)
        at async*withWrite (src/dag/lazy-store.ts:169:19)
        at async*_mutate (src/replicache.ts:1148:30)
        at async*_register/< (src/replicache.ts:1108:19)
        at src/replicache-subscribe.test.ts:911:19
```

grgbkr:
In tests add a uuid to indexeddb database names (Replicache name and IDBDatbasesStore DB) 
to isolate tests' indexed db state.
 
Also fixes a bug in kv.IDBStore which was blocking IndexedDB opened by these stores from being 
deleted.   In order to not block deletion of the db, the connection needs to be closed on 
`onversionchange`.  Previously the code was only setting up 
`db.onversionchange = () => db.close()` in `onupgradeneeded` which only fires if the db didnt 
already exist.  Code is updated to always setup `db.onversionchange = () => db.close()`. 

While this fix allows the IndexedDB databases to be reliably deleted, it did not prevent races 
around deletion.  Before isolating with uuid, I was observing that after one test's teardown 
await the deletion of a database with name X, if the next test opened the database with name 
X, the test would _sometimes_ get an error that its connection to X was closed, suggestion some 
race where the deletion is not truly complete when the success callback for the deletion is invoked.

Fixes #819


perf/replicache.ts
```diff
@@ -13,23 +13,35 @@ import {
...
 }): Benchmark {
+  let repToClose: Replicache | undefined;
   return {
     name: `populate ${valSize}x${opts.numKeys} (${
       opts.clean ? 'clean' : 'dirty'
     }, ${`indexes: ${opts.indexes || 0}`})`,
     group: 'replicache',
     byteSize: opts.numKeys * valSize,
+    async setupEach() {
+      setupIDBDatabasesStoreForTest();
```
arv, line 37: It seems like all of these tests do the same thing in setup and tear down. I guess it is non trivial to refactor to share code but maybe worth thinking about?

perf/replicache.ts
```diff
@@ -452,9 +480,16 @@ async function makeRepWithPopulate() {
...
+  if (rep) {
+    await rep.close();
+    await kv.dropIDBStore(rep.idbName);
+  }
+}
+
 export function benchmarks(): Benchmark[] {
   return [
-    // write/sub/read 1mb
+    // // write/sub/read 1mb
```
arv: //

arv (approved):
LGTM

perf/replicache.ts
grgbkr: They do similar things but not the same.  We have 5 different benchmarks, and actually 3 different setup/teardown processes, for these different setups:
1. shared rep across all runs
2. shared rep across all runs relying on shared persisted state
3. rep per run

perf/replicache.ts
grgbkr: done

PR #828, opened 2022-02-08
fix: Fix flakiness of replicache-persist.test.ts on Webkit

grgbkr:
**Problem**
replicache-persist.test is flacky on webkit because the persist process does not always complete before we 
create a new replicache and try to read the persisted data.  This happens more on webkit because it uses a 
timeout rather than request idle callback to start persist.

**Solution**
Wait for persist to complete (detected by polling the ClientMap) before creating a new Replicache and verifying
it bootstraps from the persisted data.

### DIFF
--- a/src/replicache-persist.test.ts
+++ b/src/replicache-persist.test.ts
@@ -9,14 +9,36 @@ import {expect} from '@esm-bundle/chai';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-expect-error
 import fetchMock from 'fetch-mock/esm/client';
+import * as kv from './kv/mod';
+import * as dag from './dag/mod';
+import * as persist from './persist/mod';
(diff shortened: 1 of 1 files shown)


src/replicache-persist.test.ts
```diff
@@ -34,10 +56,31 @@ test('basic persist & load', async () => {
...
 
-  // At least PERSIST_TIMEOUT should have passed.
-  await tickAFewTimes(10, 100);
+  // maxWaitAttempts * waitMs should be at least PERSIST_TIMEOUT
+  // plus some buffer for the persist process to complete
+  const maxWaitAttempts = 20;
+  const waitMs = 100;
+  let waitAttempt = 0;
+  const run = true;
+  while (run) {
```
arv, line 68: Heh.

arv (approved):

PR #829, opened 2022-02-08
add licensing client

phritz:


phritz:
@arv feel free to merge this after you add aaron and greg to the npm team

tool/get-license.ts
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env node
+
+// If developing Replicache locally, run `npm link` to link the
+// command. Then you can run this command with `npx get-license`.
+// You should just be able to run `npx get-license` if you are
+// in a package that has Replicache as a local depenency. (Note:
+// if not, the binary lives under node_modules/.bin/, we would
+// need to run it from there instead.)
+
+import * as licensingCLI from '@rocicorp/licensing/src/cli';
```
arv, line 10: import {main} from ...

arv (approved):

tool/get-license.ts
phritz: Question about this: I commonly see people import things from packages like that `import {foo} from bar`. However often I *like* to refer to the package bar when I use foo so I know what it is. Like if I'm just using `Request` and `Response` I sometimes forget what they are (http? a json type?), but it's always clear to me if I `import * as express from "express"` and then use them as `express.Request` and `express.Response`. Is this not done / bad? Or is the bad thing I'm doing importing *?

arv: They are equivalent.

It is mostly a personal style. You can always look at the import to see what it means and it should be obvious from the context.

PR #830, opened 2022-02-09
refactor!: Rename ReplicacheOptions.name and Replicache.name to userID.

grgbkr:
**Problem**
It is important, but not obvious that `ReplicacheOptions.name` uniquely identifies the user, authenticates the same user as `ReplicacheOptions.auth`, and that `ReplicacheOptions.auth` of a Replicache instance can be used to push the mutations of other Replicache instances with the same name.

**Solution**
Rename `ReplicacheOptions.name` to `ReplicacheOptions.userID`, and document its uses. 
Deprecate `Replicache.name` in favor of `Replicache.userID`.

BREAKING CHANGE: Use of `ReplicacheOptions.name` must be replaced with `ReplicacheOptions.userID`.

### DIFF
--- a/src/replicache-options.ts
+++ b/src/replicache-options.ts
@@ -37,14 +37,20 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
   pullURL?: string;
 
   /**
-   * The name of the Replicache database.
+   * A unique identifier for the user authenticated by
+   * [[ReplicacheOptions.auth]]. Must be non-empty.
    *
-   * It is important to use user specific names so that if there are multiple
-   * tabs open for different distinct users the data is kept seperate.
+   * This is used to keep different user's state separate.
    *
-   * You can use multiple Replicache instances as long as the names are unique.
+   * For efficiency, a new Replicache instance will initialize its state from
+   * the persisted state of an existing Replicache instance with the same
+   * `userID`, domain and browser profile.
+   *
+   * Mutations from one Replicache instance may be pushed using the
+   * [[ReplicacheOptions.auth]] of another Replicache instance with the same
+   * `userID`, domain and browser profile.
    */
-  name: string;
+  userID: string;
 
   /**
    * The schema version of the data understood by this application. This enables
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -68,8 +68,8 @@ export type MaybePromise<T> = T | Promise<T>;
 
 type ToPromise<P> = P extends Promise<unknown> ? P : Promise<P>;
 
-export function makeIdbName(name: string, schemaVersion?: string): string {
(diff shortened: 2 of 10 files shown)


src/replicache-options.ts
```diff
@@ -37,14 +37,20 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
...
-   * You can use multiple Replicache instances as long as the names are unique.
+   * For efficiency, a new Replicache instance will initialize its state from
+   * the persisted state of an existing Replicache instance with the same
+   * `userID`, domain and browser profile.
+   *
+   * Mutations from one Replicache instance may be pushed using the
+   * [[ReplicacheOptions.auth]] of another Replicache instance with the same
+   * `userID`, domain and browser profile.
    */
-  name: string;
```
grgbkr, line 47: I deleted rather than deprecating ReplicacheOptions.name because I want to make userID required, in which case it doesnt make sense to still have a name option.

I deprecated Replicache.name, and assigned it the value of `userID`.

What do you think @arv ?

arv: I was thinking be`name` would just be an alias to `userID` for now.

src/replicache.ts
```diff
@@ -145,15 +145,21 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
-  /** The name of the Replicache database. */
-  readonly name: string;
+  /** A unique identifier for the user authenticated by [[auth]]. */
+  readonly userID: string;
+
+  /**
+   * The name of the Replicache database.
+   * @deprecated Replaced by [[userID]].
+   */
+  readonly name?: string;
```
arv, line 156: Maybe change this to a getter returning userID?

arv (commented):
Can you also go through docs/?

A few things worth making clear in the jsdocs/docs is that userID is not sent to the server and does not have to match whatever concept of user id the server uses. It just needs to be unique. It also needs to be unique even if you do not use any kind of auth to ensure that different users are not using each other's local data.

aboodman (approved):

src/replicache-options.ts
aboodman: I prefer to just make it a breaking change cleanly. 9.0 is already a breaking version and it's important people understand the semantics of this field.

aboodman:
Nit: @arv has in the past pestered me to use my own fork for PR branches. Forwarding the pester (assuming we are still doing that).

aboodman:
Note: I pushed a new version of this change fixing the conflict.

arv:
> Nit: @arv has in the past pestered me to use my own fork for PR branches. Forwarding the pester (assuming we are still doing that).

That is the common way but I think now that we depend on secrets in GH actions I do not think that works. Given that, putting branches on rocicorp/replicache is unavoidable, but please name them `username/description` and delete them when done.

PR #831, opened 2022-02-09
add npmrc setup to perf CI

phritz:
The perf CI previously passed -- how did that even work? Maybe it got lucky and re-used a workspace?


phritz:
Wat. The checks showed green again but the perf test was broken?

PR #835, opened 2022-02-10
refactor: Rename ReplicacheOptions.userID and Replicache.userID back to name.

grgbkr:
This reverts commit d95d8bf18d84dc465f2cbe58069668fc86a2a6cd.

We realized we will need both a userID, and another identifier to support multiple Replicache instance for the same user (e.g. roomID).  We will do this api change in v10 rather than v9.  

Added details to documentation for `name` around Replicache bootsrapping and mutation recovery.


aboodman (approved):
Will review docs for stable release.

PR #846, opened 2022-02-24
fix: Catch errors in background interval processes and log appropriatly.

### LINKED ISSUE #843: Database is closing error in heartbeat
It is common to close and reopen a db quickly (in react hot module reloading for example) but that leads to issues with the db being in the process of being closed.

Before SDD we had a way to detect this by awaiting the closing promise when we open the db again. I think the `_perdag` db is not doing this as we expect any more leading to this error.

See https://discord.com/channels/830183651022471199/942807273300955187

grgbkr:
Problem
=======
We have a report of the following exception being thrown from `heartbeat.ts` after a Replicache instance is closed and a new one created as part of a development setup using Replicache, Reach hooks, and Next with HMR.

```
DOMException: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.
```

We do stop the heartbeat interval when a Replicache instance is closed, however there is a race that can lead to the above exception: if the Replicache instance is closed while the hearbeat update is running.  

This is a fairly narrow race, so I'm still uncertain if this is what the issue reporter is hitting. 

Solution
=======
Catch errors in interval based background processes and log them to 'debug' if the error occurred after the Replicache instance was closed (as this is an expected error), and to 'error' otherwise.  Applied this to the "heartbeat" and "ClientGC" processes.  The "mutation recovery" process already does this.

Also added so additional debug logging to aid in further debugging if this does not fix the issue for the reporter.

Fixes #843

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -47,13 +47,13 @@ export class LogContext implements Logger {
   readonly info?: (...args: unknown[]) => void = undefined;
   readonly error?: (...args: unknown[]) => void = undefined;
 
-  constructor(level: LogLevel = 'info', s = '') {
+  constructor(level: LogLevel = 'info', s = '', c = console) {
     this._s = s;
 
     const impl =
       (name: LogLevel) =>
       (...args: unknown[]) =>
-        console[name](this._s, ...args);
+        c[name](this._s, ...args);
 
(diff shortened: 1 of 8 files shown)


arv (commented):
LGTM

PR #848, opened 2022-02-28
WIP DO NOT MERGE attempt to integrate status check

phritz:


src/replicache.ts
```diff
@@ -302,13 +302,50 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+      licensing
+        .GetLicenseStatus(
+          f,
+          licensing.PROD_LICENSE_SERVER_URL,
+          this._licenseKey,
+        )
+        // TODO I don't think I'm supposed to be using then() for async calls
+        // but we definitely don't want to block Replicache startup here. I
+        // think using await would do that. Is there a more idiomatic
+        // way to schedule this call -- I think perhaps nextTick or similar?
```
arv, line 320: Do you really want to do this work inside the constructor? Generally the constructor should construct things... Maybe do it in open or some other async method?

src/replicache.ts
```diff
@@ -302,13 +302,50 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+      // I went down the path of expanding the type of node fetch in the lincensing
+      // client to accommodate the browser's type, but it's a neverending and increasingly
+      // complicated set of type extensions (see that PR).
+      //
+      // Q:
+      // 1. Will this work?
+      // 2. What's an easy way for me to test this in a real browser?
+      // 3. Should I switch strategy and try to constructively define in the licensing
+      //    client type the subset that we actually use?
+      const f = fetch as licensing.FetchFunc; // <== suspect
```
arv: Why do you even need to declare the type? I assume it is because you want to pass the function into `getLicenseStatus`? How is that declared? Is that declared as:

```ts
async function getLicenseStatus(f: typeof fetch, ...) {
```

src/replicache.ts
phritz: cc @aboodman 
1. Good point, guess I can do it on the first open, and ensure it only runs once. Yes?
2. What's the mechanism I should use to kick the call off in the background? I have then() here but I think I should use something like nextTick()?

phritz: Re 1, open actually happens in the constructor so can't do it there. Maybe when they get a tx, looking.

grgbkr: 1.  We call this._open at the end of this constructor.  The constructor and this._open always run the same number of times.  
2. One pattern we've used for background things like this is to put them in an async function, and then call them prefixed with void (which tells our linter that we are intentionally not handling the promise).  The call to this._open at the end of this constructor is an example of this.  It looks like `void this._open(clientIDResolver.resolve, readyResolver.resolve);`.

phritz:
sorted, thanks guys

PR #849, opened 2022-03-01
add license check call

phritz:


src/replicache.ts
```diff
@@ -404,6 +399,41 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+      this._logger.info?.(
+        `Skipping license check for TEST_LICENSE_KEY. ` +
+          `You may ONLY use this key for automated (e.g., unit/CI) testing. ` +
+          // TODO(phritz) maybe use a more specific URL
+          `See https://replicache.dev for more information.`,
+      );
+      return true;
+    }
+    try {
+      const status = await licensing.GetLicenseStatus(
```
aboodman, line 418: camel case

src/replicache.test.ts
```diff
@@ -24,6 +24,7 @@ import {WriteTransactionImpl} from './transactions';
 import {emptyHash, Hash} from './hash';
 import {defaultPuller} from './puller';
 import {defaultPusher} from './pusher';
+import * as licensing from '@rocicorp/licensing/src/client';
```
aboodman, line 27: It's more conventional to directly import the names used, like:

```
import { PROD_LICENSE_SERVER_URL, TEST_LICENSE_KEY, LicenseStatus, getLicenseStatus} from ...
```

In the case where the imported names are overly generic or conflict with some other name, they can be renamed like:

```
import {foo as bar} from ....
```

src/fetch-wrapper.ts
```diff
@@ -0,0 +1,20 @@
+import type {FetchFunc, FetchResponse} from '@rocicorp/licensing/src/client';
+
+export const fetchWrapper: FetchFunc = browserFetchWrapper;
```
aboodman: This all works but it doesn't feel idiomatic to me... here are some thoughts:

It's not common to specify the type of a function this way -- their types are implicit from the argument and return types.  We don't say:

```ts
type Sum = (a: number: b: number) => number;
function sum: Sum(a: number, b: number): number { ... }
```

... we just say the second line, the type is implied.

This is commonly done even when the desired type is a subset of the actual type of the function, as in your case. Structural typing to the rescue:

```ts
function doStuff(logger: (msg: string) => void) {
  ...
}

function consoleLog(msg: string, andCrash?: boolean|undefined) {
  ...
}

// No need to explicitly say that consoleLog is typeof `logger` -- if it isn't this will fail at compile time
doStuff(consoleLog)
```

So for your case, it would be more common to say:

```
export async function browserFetchWrapper(method: string, url: string, body: string | null, headers: string[][]): Promise<Response> {
    ...
    return fetch(url, responseInit);
}
```

Any attempt to pass this function to a var with type `FetchFunc` will fail at compile time unless `browserFetchWrapper` actually is a superset of `FetchFunc`.

Second, I would suggest a more descriptive name for this type. How about `simpleFetch`? I'd define the type `SimpleFetch` inside `licensing/src/client` and then implement it here (and presumably in the node version) like this:

```
export function async browserSimpleFetch(method: string, url: string, body: string | null, headers: string[][]): Promise<FetchResponse> {
...
}
```

The caller can then pass this `browserSimpleFetch` wherever an impl of `SimpleFetch` is wanted. There's no need for both impls to export the same name `simpleFetch` or `fetchWrapper` since call site will be specific to one implementation and will be passing the variable to some common function.

aboodman (commented):
this review might not entirely make sense i'm half asleep and running to a meeting so can't double-check.

hope it's helpful still.

src/replicache.ts
phritz: yeah you pointed that out on the other review, i plan to fix next time i touch the code where these functions are defined (its in a separate repo)

src/replicache.test.ts
phritz: In a file with lot of imports I *like* having the name of the module prefix the thing I'm using, like `licensing.foo` is a good reminder where foo comes from and what it is. Is that not something people do? By your comment guessing no...

src/fetch-wrapper.ts
phritz: Gotcha on the typing. This was my attempt to do a go-like type assertion so it's clear what the intent was. I do see how it is not idiomatic and also kind of useless, will fix.

Re name, will update the name next time I touch that repo (later today).

phritz:
thanks @aboodman, merged, will make the changes to client today

src/simple-fetch.ts
```diff
@@ -0,0 +1,18 @@
+import type {FetchResponse} from '@rocicorp/licensing/src/client';
+
+// This wrapper around fetch adapts it to the narrow fetch interface
+// that the licensing client expects. That type enables the licensing
+// client to work with both the browser and node-fetch implementations.
+export async function browserSimpleFetch(
```
arv, line 6: Why not just narrow the type instead of having a runtime wrapper??

If you really want to be this precise I recommend defining the following type:

```ts
type SimpleResponse = {
  status: number;
  text: () => Promise<string>;
};

type SimpleFetch = (
  url: string,
  options: {
    method: string;
    body: string | null;
    headers: string[][];
  }
) => Promise<SimpleResponse>;
```

(or all it FetchSubset or something like that)

Then you can directly pass `fetch` and `nodeFetch` without a runtime wrapper

aboodman: My understanding was that there was no useful common subset of fetch between the browser and node ones, hence the need for the adaptor.

src/replicache.test.ts
aboodman: Yeah it's not commonly done. I guess because the ability to rename identifiers at import time addresses the concern of possibly ambiguous names without imposing the cost of `licensing.` on names that are already unambiguous.

src/simple-fetch.ts
arv: The above types cover the usages and there is nothing in the above type that is would expose those underlying differences.

PR #850, opened 2022-03-01
move license check to after ready

phritz:
Kicking the license check off before ready slows down startup on perf tests even though it's a nop.

https://github.com/rocicorp/replicache/actions/runs/1918318439

cc @grgbkr


phritz:
I'm having trouble reproducing locally, will merge and see what it tells me.

PR #852, opened 2022-03-04
Add 64MB write/sub/read benchmark 😬

aboodman:
This setup for this benchmark is slow but I figure we need it so we can make it faster separately.


grgbkr:
Nice.  LGTM.

PR #853, opened 2022-03-04
Aa/benchmark perf improvement

aboodman:


perf/data.ts
```diff
@@ -50,13 +50,15 @@ export function randomASCIIString(len: number): string {
   return randomStringInternal(len, 128);
 }
 
+// Shared array buffer to reduce allocations.
+const codes: number[] = [];
```
aboodman, line 54: In my measurements this function was taking ~30% of execution time on the 16MB benchmark, mostly due to GC. Sharing this buffer reduced the time to ~15%.

perf/replicache.ts
```diff
@@ -359,16 +359,15 @@ export function benchmarkWriteSubRead(opts: {
     async run(bencher: Bencher) {
       const keys = Array.from({length: numKeys}, (_, index) => makeKey(index));
       const sortedKeys = keys.sort();
-      const data: Map<string, TestDataObject> = new Map(
-        keys.map(key => [key, jsonObjectTestData(valueSize)]),
-      );
+      const data: Map<string, TestDataObject> = new Map();
```
aboodman, line 362: My theory here was that holding onto these random objects for less time would reduce GC pressure, but I'm not 100% sure it did anything.

grgbkr: They aren't held for much less time right?  in init below, we put them into replicache, which will keep a reference to them for the full test.

perf/replicache.ts
```diff
@@ -359,16 +359,15 @@ export function benchmarkWriteSubRead(opts: {
...
 
+      const t0 = Date.now();
```
grgbkr, line 364: remove

perf/replicache.ts
aboodman: Ah good point, forgot about that. I'll abandon this one.

PR #856, opened 2022-03-04
Minor perf optimization to benchmarks.

aboodman:
This reduces time of the 16MB run by about 10%. It has no affect on the result of the benchmark, just the setup overhead.


aboodman:
Sorry I'm not sure why I didn't reuse the old PR :).

PR #864, opened 2022-03-10
fix: Log subscribe errors if no onError

### LINKED ISSUE #862: Do we really swallow subscriptions that throw by default??
I just did `useSubscribe(...thing that throws...)` and ... nothing showed up anywhere. What the? It looks like we swallow errors from subscription functions.

We should send them to `console.error()` by default.

arv:
If `onError` is provided to a subscription the `onError` handler gets
called with the exception as argument.

If there is no `onError` then we log the error to the console.

Fixes #862


aboodman (commented):
Thanks!

Issue #784, opened 2022-01-17
Error handling for old tab that is visited after 1 week and discovers its client state has been GC'd


arv:

arv:
We talked about this and the idea is to have a callback, lets call it `onClientDisappeared` (strawman). The default implementation would be to do a `location.reload()` but allow the API would allow you to override this to do something else.

arv:
We should change the type signature in a way that we cannot forget to check if the perdag went away...

arv:
What happens is that the client head goes away in the `perdag` which results in chunks being garbage collected. 
- When the `LazyStore` tries to read data it will get missing chunks if the chunks have not been cached.
- However, the `LazyStore` might have the chunks cached and there is no telling **when** we get an error. It will happen if we ever have to evict chunks from the `memdag`; this can be a long time in the future.
- When we do a `persist` of a collected client we create a new client entry in the clients head which blurs the picture even further.

Ways to detect that the client has been collected:
- We get missing chunk errors when reading from the perdag.
- There is no client in the clients map

To make things further robust, if we get a missing chunk error we then check if the client exists in the clients map. If it doesn't we know the client was garbage collected.

aboodman:
Sounds great. No matter what if there is really a missing chunk the client is hosed, but it's the difference between an internal error and an expected error due to GC. If a gc-related issue a reload fixes it and we should default to that as described above, but if it's an internal issue we should just throw for now or print to console. Perhaps in the future we should also allow developer to hook this as discussed in #462 .

aboodman:
One thought: since the client can only really get GC'd when the tab is switched away from, what if we just check that the client still exists on focus and if it doesn't raise the event to developer.

I guess it would still be good for correctness to still have the `mustGetChunk()` feature, but I guess what I'm asking is: should we raise the event to dev by catching this `MissingChunk` error or should we raise by checking for missing client in focus. Seems simpler to check in focus and since this is a path that will be very rarely run simple seems good. WDYT?

arv:
`focus` -- if we do that we should at least use `pageshow` instead.

PR #866, opened 2022-03-11
chore: Add a mustGetChunk that throws if missing

arv:
Now, all code paths that gets a required chunk uses `mustGetChunk`. When
the chunk is missing this throws a `MissingChunkError`.

The idea is that the caller will detect these errors and see if the
client might have been GC'd.

Towards #784


grgbkr:
LGTM.

PR #867, opened 2022-03-11
feat: Check if client exists in persist

arv:
We now check if the client ID exists in the client map when we do a
`persist`. If it doesn't we throw a `MissingClientError`.

For testing purpuse we can skip this check.

The intended use is to handle clients that are missing and raise an
"event" on the Replicache instance when this happens.

Towards #784

### DIFF
--- a/src/persist/persist.ts
+++ b/src/persist/persist.ts
@@ -4,7 +4,7 @@ import * as db from '../db/mod';
 import * as sync from '../sync/mod';
 import {Hash, hashOf} from '../hash';
 import type {ClientID} from '../sync/client-id';
-import {assertclientExists, updateClients} from './clients';
+import {assertClientExists, updateClients} from './clients';
 import {ComputeHashTransformer, FixedChunks} from './compute-hash-transformer';
 import {GatherVisitor} from './gather-visitor';
 import {FixupTransformer} from './fixup-transformer';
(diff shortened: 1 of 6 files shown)


src/replicache.test.ts
```diff
@@ -2165,7 +2165,7 @@ test('experiment KV Store', async () => {
   store.resetCounters();
 
   await rep.persist();
-  expect(store.readCount).to.equal(1, 'readCount');
+  expect(store.readCount).to.equal(2, 'readCount');
```
arv, line 2168: One more read because we read to check for the clients...

Maybe we can bake this into the write transaction. Let me look into that.

arv: > Maybe we can bake this into the write transaction. Let me look into that.

No. Then we have to bake that into `updateClients` which seems too strange,

src/persist/persist.ts
```diff
@@ -25,21 +25,29 @@ export async function persist(
   clientID: ClientID,
   memdag: dag.Store,
   perdag: dag.Store,
+  skipClientExists?: 'skip',
```
grgbkr, line 28: I think its best not to skip this in tests.  

It's straightforward to set up the expected client using the test helpers in `clients-test-helpers.ts`.

src/replicache.test.ts
grgbkr: This tests is a bit annoying.

grgbkr:
LGTM

src/persist/persist.ts
arv: Will do.

src/replicache.test.ts
arv: Agreed. Maybe change the test to check what is in the experimental kv store instead of counting the reads/writes/close?

PR #868, opened 2022-03-12
feat: Add onClientMissing

### LINKED ISSUE #784: Error handling for old tab that is visited after 1 week and discovers its client state has been GC'd

arv:
This hooks up the test if the client exists. If the client does not
exist we call `onClientMissing`.

The test for the client missing is done in persist, query, mutate as well as in
visibilitychange and the visibilityState is visible.

Fixes #784

### DIFF
--- a/src/persist/client-gc.ts
+++ b/src/persist/client-gc.ts
@@ -47,3 +47,16 @@ export async function gcClients(
     };
   }, dagStore);
 }
+
+export async function deleteClientForTesting(
+  clientID: ClientID,
+  dagStore: dag.Store,
+): Promise<void> {
+  await updateClients(clients => {
+    const clientsAfterGC = new Map(clients);
+    clientsAfterGC.delete(clientID);
+    return {
+      clients: new Map(clientsAfterGC),
+    };
+  }, dagStore);
+}
--- a/src/persist/client-gc.ts
+++ b/src/persist/client-gc.ts
@@ -47,16 +47,3 @@ export async function gcClients(
     };
   }, dagStore);
 }
-
-export async function deleteClientForTesting(
-  clientID: ClientID,
-  dagStore: dag.Store,
-): Promise<void> {
-  await updateClients(clients => {
-    const clientsAfterGC = new Map(clients);
-    clientsAfterGC.delete(clientID);
-    return {
-      clients: new Map(clientsAfterGC),
-    };
-  }, dagStore);
-}
--- a/src/persist/clients.ts
+++ b/src/persist/clients.ts
@@ -131,10 +131,13 @@ export async function assertClientExists(
   id: sync.ClientID,
   dagRead: dag.Read,
 ): Promise<void> {
-  const client = await getClient(id, dagRead);
-  if (!client) {
+  if (await isClientMissing(id, dagRead)) {
     throw new MissingClientError(id);
   }
+}
+
+export async function isClientMissing(id: sync.ClientID, dagRead: dag.Read) {
+  return !(await getClient(id, dagRead));
 }
 
 export async function getClient(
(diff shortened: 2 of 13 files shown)


src/persist/clients.ts
```diff
@@ -131,12 +131,15 @@ export async function assertClientExists(
   id: sync.ClientID,
   dagRead: dag.Read,
 ): Promise<void> {
-  const client = await getClient(id, dagRead);
-  if (!client) {
+  if (await isClientMissing(id, dagRead)) {
```
arv: Maybe we should just have persist return a boolean instead of throwing the error?

arv (commented):
Needs tests too...

I still think it would be good to do this check in the case where we get a MissingChunkError.

arv:
Tests are in place

@aboodman @grgbkr Is `onClientMissing` the best name?

src/replicache.ts
```diff
@@ -277,6 +277,16 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
 
+  /**
+   * `onClientMissing` is called when the persistent client has been garbage
+   * collected. This can happen if the client has not been used for over a week.
+   *
+   * The default behavior is to reload the page (using `location.reload()`). Set
+   * this to `null` or provide your own function to prevent the page from
+   * reloading automatically.
+   */
+  onClientMissing: (() => void) | null = reload;
```
aboodman: Maybe we should only assign the `reload` in the case we're in a document so at least workers get an error?

aboodman: How about `onClientMissingError`.

aboodman: No great inspiration on the name though --- @grgbkr ?

grgbkr: Perhaps `onClientStateNotFound` and `MissingClientError` becomes `ClientStateNotFoundError`.  I don't feel strongly about it though.

src/persist/client-gc.ts
```diff
@@ -47,3 +47,16 @@ export async function gcClients(
     };
   }, dagStore);
 }
+
+export async function deleteClientForTesting(
```
grgbkr: This should go in `clients-test-helpers.ts`.

src/replicache.ts
```diff
@@ -1046,9 +1090,21 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+    } catch (e) {
+      if (e instanceof persist.MissingClientError) {
+        this._fireOnClientMissing();
+      } else {
+        throw e;
+      }
+    }
+  }
+  private _fireOnClientMissing() {
+    this._logger.error?.('Client is missing');
```
grgbkr: suggest also logging client id.

grgbkr (commented):
Thanks for doing this error handling.  Looks great.  One other place where "missing client state" comes up is in heartbeat.   Checkout `heartbeat.ts`, perhaps that should also be wired up to `onClientMissing`?

src/replicache.ts
arv: > Maybe we should only assign the reload in the case we're in a document so at least workers get an error?

reload is only done if there is a `location` global. We could do the check in the constructor but this seems simpler.

src/persist/heartbeat.ts
```diff
@@ -1,25 +1,41 @@
...
   lc: LogContext,
 ): () => void {
   return initBgIntervalProcess(
     'Heartbeat',
-    () => {
+    async () => {
       latestHeartbeatUpdate = writeHeartbeat(clientID, dagStore);
-      return latestHeartbeatUpdate;
+      try {
+        return await latestHeartbeatUpdate;
```
arv, line 27: @grgbkr The Promise was never awaited before...

src/replicache.ts
aboodman: Not sure if we're saying the same thing. What I'm saying is that it's desirable that if a worker encounters a `ClientNotFoundError` that the error actually get reported to the console, not that it's silently swallowed.

src/persist/heartbeat.ts
grgbkr: It was returned and awaited in `initBgIntervalProcess`

PR #869, opened 2022-03-14
chore: Add another scan test with ~10MB

arv:


aboodman:
OK, it's exactly linear. Interesting thanks!

arv:
> OK, it's exactly linear. Interesting thanks!

That is what I expected

PR #870, opened 2022-03-14
Performance results

arv:
This visualizes performance runs on M1 (2020) and MacPro (2015)

It runs the perf tests with 4kb and 8kb chunk size.

The p95 results are removed from the graph.

Lower is better

### DIFF
--- a/perf-results.js
+++ b/perf-results.js
@@ -1,0 +1,2930 @@
+const m1_min_8kb_max_16kb_run1 = [
+  {
+    name: 'writeSubRead 1MB total, 64 subs total, 5 subs dirty, 16kb read per sub',
+    unit: 'median ms',
+    value: 0.7999999970197678,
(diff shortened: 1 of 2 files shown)


arv:
You can run this using `npx http-sever` or any other static file http server

Lower is better

Screen Shot 2022-03-14 at 17 11 22

It looks like we should stick with 8kb

arv:
writeSubReadOnly

Screen Shot 2022-03-14 at 17 15 05

arv:
scan onlye

Screen Shot 2022-03-14 at 17 16 00

arv:
@aboodman @grgbkr

arv:
I moved this to codesandbox.

https://codesandbox.io/s/repl-btree-chunk-size-wnp5vy?file=/src/index.js
https://wnp5vy.csb.app/

PR #875, opened 2022-03-17
chore: Remove flag to persist

arv:
Fix tests to not need to skip the checking of missing clients

Followup to #867


grgbkr:
Thanks. LGTM.

PR #880, opened 2022-03-18
feat: New Phone, Who Dis?

### LINKED ISSUE #335: new phone who dis
It is frequently the case, especially during development, but occasionally in production too, that a server just wants to drop a particular client's state. Just... forget it. rm -rf, start over.

Unfortunately, when that happens, the client is then wedged. It goes to sync, uses its existing client ID, and the data layer doesn't recognize them.

We need a way for the server and client to negotiate this, and end up with the client getting a fresh state too.

arv:
The server can now tell the client that it does not know about a client.
When this happens the client calls `onClientStateNotFound`.

The server can return:

```json
{"error": "ClientStateNotFound"}
```

Fixes #335

### DIFF
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -4,7 +4,13 @@ import {Lock} from '@rocicorp/lock';
 import {deepClone, deepEqual, ReadonlyJSONValue} from './json';
 import type {JSONValue} from './json';
 import {Pusher, PushError} from './pusher';
-import {Puller, PullError, PullResponse} from './puller';
+import {
+  isClientStateNotFoundResponse,
+  Puller,
+  PullError,
+  PullResponse,
+  PullResponseOK,
+} from './puller';
 import {
   SubscriptionTransactionWrapper,
   IndexTransactionImpl,
@@ -988,6 +994,12 @@ export class Replicache<MD extends MutatorDefs = {}> {
     const lc = this._lc
       .addContext('handlePullResponse')
       .addContext('request_id', requestID);
+
+    if (isClientStateNotFoundResponse(poke.pullResponse)) {
+      this._fireOnClientStateNotFound(clientID);
+      return;
+    }
+
     const syncHead = await sync.handlePullResponse(
       lc,
       this._memdag,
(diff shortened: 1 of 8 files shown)


aboodman:
My only API feedback is perhaps we should add a parameter to `onClientStateNotFound` that indicates whether it was the client-side or server-side state which was missing.

arv:
> My only API feedback is perhaps we should add a parameter to `onClientStateNotFound` that indicates whether it was the client-side or server-side state which was missing.

I looked that. I'll add a reason parameter

src/replicache.ts
```diff
@@ -1599,30 +1617,49 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+            lastMutationID: pullResponse?.lastMutationID,
+          },
+        );
+      }
       const newClientMap = await persist.updateClients(
         (clients: persist.ClientMap) => {
           assertNotUndefined(pullResponse);
+
           const clientToUpdate = clients.get(clientID);
+          if (isClientStateNotFoundResponse(pullResponse)) {
```
grgbkr:
```suggestion
  if (!clientToUpdate) {
    return persist.noClientUpdates;
  }
  if (isClientStateNotFoundResponse(pullResponse)) {
    const newClients = new Map(clients);
    newClients.delete(clientID);
    return {clients: newClients};
  }
  if (clientToUpdate.lastServerAckdMutationID >=
      (pullResponse as PullResponseOK).lastMutationID) {
    return persist.noClientUpdates;
  } 
```

src/replicache.ts
```diff
@@ -1599,30 +1617,49 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+            if (!clientToUpdate) {
+              return persist.noClientUpdates;
+            }
+
+            const newClients = new Map(clients);
```
grgbkr, line 1631: This is basically an early GC of the client.  I guess this is the best thing to do as these mutations will never be successfully recovered.  Wondering if we should log about this at some higher level than debug.

grgbkr:
LGTM

PR #890, opened 2022-03-22
chore!: Removal of deprecated Replicache methods.

arv:
This removes the following deprecated methods from the Replicache
instance:
- scan
- has
- isEmpty
- get

These have been deprecated since 8.0


grgbkr:
Thanks for doing this clean up.  LGTM

PR #903, opened 2022-03-29
enable license check enforcement by default

phritz:
also includes a few small cleanups requested by aa

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "module": "out/replicache.mjs",
   "types": "out/replicache.d.ts",
   "bin": {
-    "get-license": "./out/get-license.cjs"
+    "get-replicache-license": "./out/get-license.cjs"
   },
   "exports": {
     "module": "./out/replicache.mjs",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "devDependencies": {
     "@esm-bundle/chai": "^4.3.4",
-    "@rocicorp/licensing": "^4.0.0-beta.2",
+    "@rocicorp/licensing": "^4.0.0-beta.3",
     "@rocicorp/lock": "^1.0.1",
     "@rocicorp/logger": "^2.1.1",
     "@rocicorp/resolver": "^1.0.0",
--- a/src/replicache-options.ts
+++ b/src/replicache-options.ts
@@ -178,19 +178,22 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
   pusher?: Pusher;
 
   /**
+   * The license key for Replicache. This parameter is required for Replicache to
+   * function. See https://replicache.dev for how to acquire a license key.
+   *
+   * YOU SHOULD PASS Replicache.TEST_LICENSE_KEY IN AUTOMATED TESTS. It disables license
+   * checks for several minutes. If you pass a normal license key in tests, each test
+   * that instantiates Replicache will attempt to perform a license check against
+   * Replicache's licensing server, potentially increasing your monthly active browser
+   * profile count, slowing the test down, and spamming Replicache's servers.
+   */
+  licenseKey?: string;
+
+  /**
    * Allows implementing the underlying storage layer completely in JavaScript.
    *
    * @experimental This option is experimental and might be removed or changed
    * in the future without following semver versioning. Please be cautious.
    */
   experimentalKVStore?: kv.Store;
-
-  /**
-   * When defined, activates Replicache's experimental license key support. You
-   * don't want to use this option unless you are a Replicache developer.
-   *
-   * @experimental This option is experimental and might be removed or changed
-   * in the future without following semver versioning. Please be cautious.
-   */
-  experimentalLicenseKey?: string;
 }
--- a/src/replicache.ts
+++ b/src/replicache.ts
@@ -51,11 +51,13 @@ import {
   getLicenseStatus,
   licenseActive,
   PROD_LICENSE_SERVER_URL,
-  TEST_LICENSE_KEY,
   LicenseStatus,
+  TEST_LICENSE_KEY as TLK,
 } from '@rocicorp/licensing/src/client';
 import {mustSimpleFetch} from './simple-fetch';
 import {initBgIntervalProcess} from './persist/bg-interval';
+
+export const TEST_LICENSE_KEY = TLK;
 
 export type BeginPullResult = {
   requestID: string;
@@ -210,7 +212,7 @@ export class Replicache<MD extends MutatorDefs = {}> {
 
   /* The license is active if we have sent at least one license active ping
    * (and we will continue to). We do not send license active pings when
-   * licensing is disabled (license key is undefined) or for the TEST_LICENSE_KEY.
+   * for the TEST_LICENSE_KEY.
    */
   protected _licenseActivePromise: Promise<boolean>;
   private _stopLicenseActive = noop;
@@ -337,8 +339,8 @@ export class Replicache<MD extends MutatorDefs = {}> {
       requestOptions = {},
       puller = defaultPuller,
       pusher = defaultPusher,
+      licenseKey,
       experimentalKVStore,
-      experimentalLicenseKey,
     } = options;
     this.auth = auth ?? '';
     this.pullURL = pullURL;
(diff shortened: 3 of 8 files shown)


src/replicache-options.ts
```diff
@@ -178,19 +178,22 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
...
+   * function. See https://replicache.dev for how to acquire a license key.
+   *
+   * YOU SHOULD PASS Replicache.TEST_LICENSE_KEY IN AUTOMATED TESTS. It disables license
+   * checks for several minutes. If you pass a normal license key in tests, each test
+   * that instantiates Replicache will attempt to perform a license check against
+   * Replicache's licensing server, potentially increasing your monthly active browser
+   * profile count, slowing the test down, and spamming Replicache's servers.
    */
-  experimentalKVStore?: kv.Store;
+  licenseKey?: string;
```
phritz, line 190: Every property in this interface is optional so I made this one optional as well to just follow the pattern...

package.json
```diff
@@ -58,7 +58,7 @@
   "module": "out/replicache.mjs",
   "types": "out/replicache.d.ts",
   "bin": {
-    "get-license": "./out/get-license.cjs"
+    "get-replicache-license": "./out/get-license.cjs"
```
arv, line 61: I know this is just "bike shedding" but I would prefer if this was:

```
"bin": {
  "replicache": "./out/cli.js"
}
```

because then one can do:

```
npx replicache get-license
```

If the name of the bin is the same as the npm package you can run this from anywhere and npm will download it as needed.

For example you can do:

```
npx esbuild --help
```

or

```
npx playwright help install-deps
```

from anywhere

src/replicache-options.ts
arv: FWIW, `name` is required.

I think it is OK to keep this optional and default to the `TEST_LICENSE_KEY`

src/replicache.ts
```diff
@@ -51,12 +51,14 @@ import {
...
   licenseActive,
   PROD_LICENSE_SERVER_URL,
-  TEST_LICENSE_KEY,
   LicenseStatus,
+  TEST_LICENSE_KEY as TLK,
 } from '@rocicorp/licensing/src/client';
 import {mustSimpleFetch} from './simple-fetch';
 import {initBgIntervalProcess} from './persist/bg-interval';
 
+export const TEST_LICENSE_KEY = TLK;
```
arv, line 60: I guess what you want is:

```js
import {TEST_LICENSE_KEY} from '@rocicorp/licensing/src/client';

export {TEST_LICENSE_KEY};
```

src/replicache.ts
```diff
@@ -508,12 +510,15 @@ export class Replicache<MD extends MutatorDefs = {}> {
   private async _licenseCheck(
     resolveLicenseCheck: (valid: boolean) => void,
   ): Promise<void> {
-    // Licensing is experimental, so if a key is not set don't do anything.
-    if (this._licenseKey === undefined) {
-      resolveLicenseCheck(true);
+    if (this._licenseKey === undefined || this._licenseKey === '') {
```
arv, line 513:
```
if (!this._licenseKey) {
```

src/replicache.ts
```diff
@@ -550,12 +554,27 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+  }
+
   private async _startLicenseActive(
     resolveLicenseActive: (valid: boolean) => void,
     lc: LogContext,
   ): Promise<() => void> {
     if (
       this._licenseKey === undefined ||
+      this._licenseKey === '' ||
       this._licenseKey === TEST_LICENSE_KEY
```
arv, line 578:
```js
if (!this._licenseKey || this._licenseKey === TEST_LICENSE_KEY) {
```

src/replicache-options.ts
```diff
@@ -178,19 +178,22 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
...
 
   /**
-   * Allows implementing the underlying storage layer completely in JavaScript.
-   *
-   * @experimental This option is experimental and might be removed or changed
-   * in the future without following semver versioning. Please be cautious.
+   * The license key for Replicache. This parameter is required for Replicache to
+   * function. See https://replicache.dev for how to acquire a license key.
+   *
+   * YOU SHOULD PASS Replicache.TEST_LICENSE_KEY IN AUTOMATED TESTS. It disables license
```
arv, line 184: `Replicache.TEST_LICENSE_KEY` is not right. It is not a static.

Suggestion

```
   * YOU SHOULD PASS `TEST_LICENSE_KEY` IN AUTOMATED TESTS. It disables license
```

arv (commented):
Please fix the type errors in `perf/` too.

It seems like we need to update the precommit tests to ensure that we have no type errors there

src/replicache-options.ts
phritz: I don't want to default to TEST_LICENSE_KEY, I want to default to telling them they need to supply a key. If we default to TEST_LICENSE_KEY then if a customer isn't aware of the license key requirement Replicache starts up just fine and works great for N minutes and then just mysteriously stops working. I think Replicache should work or not, and that you should explicitly configure the license key that you want. Less error prone that way. That is my opinion tho, obviously if folks think differently we can change it.

src/replicache.ts
phritz: Yes it is, thanks.

package.json
phritz: I'm not a big fan of repainting a shed that works just fine, esp since we haven't launched yet, but ok will make this change.

phritz:
> Please fix the type errors in perf/ too.
> It seems like we need to update the precommit tests to ensure that we have no type errors there

Yeah I closed this PR because I couldn't understand how we had errors there. Turns out VSCode 'rename symbol' renames symbols and leaves the edited files open _but unsaved_ in vscode, so I had local changes sitting around that I thought I had committed in an earlier PR. The root problem is as you say, that the stuff in perf/ is not built or tested as part of the normal development flow, so we don't hear about breakages there until after a PR is up or even merged. The PR I ended up moving forward with https://github.com/rocicorp/replicache/pull/905 fixed the perf stuff but the root problem is still there. I will fix the root problem by ensuring that perf/ tests are at least built but better run as a (non-perf unit) test as part of the `test` target.

src/replicache-options.ts
arv: Let's make required parameters non optional.

aboodman: > mysteriously stops working.

Does it print an error? I was imagining that it would. That would make it less mysterious.

aboodman: I'm not sure how I feel about having to put in the `TEST_LICENSE_KEY` constant. I get Fritz's point, want to try it on for size. If the license key is semantically required, let's mark it required.

However note that we still need to do a runtime check (as name does) because if the caller is javascript (not typescript) nothing prevents them from not passing this param.

phritz: > Does it print an error? I was imagining that it would. That would make it less mysterious.

Yes, it prints an error. But to just expand a little on the way I'm looking at it: the customer starts using Replicache and things are going well. Then it stops working. Yes, some percentage of people are going to look in the logs, but that percentage is not 100%, some are going to just going to say 'low quality, this is busted, time to move on'. And of those that DO look in the logs, not 100% are going to see the log line. There's a lot of log spew. 

Now I guess the argument works the other way too: IF we require they take some step like getting a key before using  Replicache themselves then we lose some percentage of people who just dont want to be bothered. 

I dunno how to cut it but it just seems simpler to me mental model wise to implement what we're telling them, which is that using Replicache requires that they get a license.

aboodman: I get where you're coming from. It would be super bad for someone to not receive the info that they need to get a license key, start using Replicache (either with 'test' or empty string if it was defaulted that way), then be confused as to why it randomly stops working.

We want to drive people (via docs, dx, etc) to actually get a license key -- not the reverse. The test license key should only be used for actual automated tests. For development, tire kicking, etc, people should get a license.

phritz:
>  will fix the root problem by ensuring that perf/ tests are at least built but better run as a (non-perf unit) test as part of the test target.

On second thought, I'm not really sure what happened. If I make the same kind of change I see the breakage immediately locally when i run `npm run build` -- it builds perf/replicache.ts so I see the breakage. No idea how it slipped by... maybe user (fritz) error?!

PR #904, opened 2022-03-29
enable license check enforcement by default

phritz:
also includes a few small cleanups requested by aa


phritz:
argh, having some weird branch issues, build failing with code that isn't this branch, gnnna try from a different branch

PR #906, opened 2022-03-30
feat!: Switch to use a ScanReader and expose it to the API.

### LINKED ISSUE #607: RFE: Turn the result of `ReadTransaction.scan()` into an interface
This would allow developers to implement `WriteTransaction` on the server side which is a real nice pattern for code reuse. See: https://github.com/rocicorp/replidraw/pull/56/commits/385b5d32a53ec067cadd1cb7fed508763ed71cb5#diff-b1688cc87a204ea3f0d8bd85a4c571a7b9fa74f62e731a327b59861f5efe7882R48

arv:
This changes scan to use a ScanReader instead of an async iterator.

It also exposes a function that returns a ScanResult from a ScanReader
and ScanOptions.

```ts
declare const options: ScanOptions;
declare const reader: ScanReader;
const scanResult createScanResultFromScanReader(reader, options);
```

If you are trying to implement Replicache's scan API you now only need
to write a function that creates a ScanReader.
createScanResultFromScanReader will take care of seeking to the correct
position and reading the data.

Known issues: This depends on how we encode keys for index maps but we
do not expose that so it is unlikely that people can reuse this for
index scans.

Closes #607


arv:
**main**

```
scan 1024x1000 x 976.56 MB/s ±1.8% (19 runs sampled)
scan 1024x10000 x 1252.00 MB/s ±3.4% (19 runs sampled)
startup scan 1024x100 from 1024x100000 stored x 6.34 MB/s ±42.2% (19 runs sampled)
```

**arv/scan-reader**

```
scan 1024x1000 x 1085.07 MB/s ±2.0% (19 runs sampled)
scan 1024x10000 x 1743.86 MB/s ±3.0% (19 runs sampled)
startup scan 1024x100 from 1024x100000 stored x 7.51 MB/s ±61.3% (19 runs sampled)
```

aboodman:
> This depends on how we encode keys for index maps but we do not expose that so it is unlikely that people can reuse this for index scans.

I see -- I forgot about indexes. Can we expose the two keys to the reader in a more structured way, like as two separate value (secondary and primary key), then let the reader do whatever it needs to do to implement the scan of the index? Our reader impl would know how to concat in the right way for our indexes.

arv:
> I see -- I forgot about indexes. Can we expose the two keys to the reader in a more structured way, like as two separate value (secondary and primary key), then let the reader do whatever it needs to do to implement the scan of the index? Our reader impl would know how to concat in the right way for our indexes.

We can but that means introducing a lot of object allocations that we would like to skip... but maybe it is fine because these gets allocated later anyway.

aboodman:
It seems like the protocol would be something like:

1. user tells replicache to scan index "byName" starting at secondary key = foo, primary key = bar
2. replicache asks app for a reader for index "byName"
3. replicache seeks reader to secondary key = foo, primary key = bar by passing two separate values
4. app reader impl does what it needs to do to implement that seek, maybe concat'ing/allocating
5. replicache wraps app reader in an impl of ScanResult
6. replicache calls next() repeatedly - app returns secondary key, primary key, and value as three distinct values.

Note unrelated to what we're discussing here, but I think at step 5 we don't need or want to make `ScanResult` an interface anymore, now that we have reader. That just creates more API we have to worry about not breaking, for no reason.

arv:
Exposing the entries directly had no impact on the performance.

It made things a bit messier but I'll try to clean it up and post a new patch.

PR #907, opened 2022-03-30
feature: Enable custom log handling by adding logSink to ReplicacheOptions

grgbkr:
This will enable customers to send logs to their own servers or to a cloud logging service (e.g. DataDog).

### DIFF
--- a/src/replicache-options.ts
+++ b/src/replicache-options.ts
@@ -92,8 +92,8 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
    * sent elsewhere (e.g. to a cloud logging service like DataDog) you can
    * provide an array of `LogSink`s.  Logs at or above
    * [[ReplicacheOptions.logLevel]] are sent to each of these `LogSink`s.
-   * If you would still like logs to go to the console, include
-   * [[consoleLogSink]] in the array.
+   * If you would still like logs to go to the console, include `consoleLogSink`
+   * in the array.
    *
    * ```ts
    * logSinks: [consoleLogSink, myCloudLogSink],
--- a/src/replicache-options.ts
+++ b/src/replicache-options.ts
@@ -92,8 +92,8 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
    * sent elsewhere (e.g. to a cloud logging service like DataDog) you can
    * provide an array of `LogSink`s.  Logs at or above
    * [[ReplicacheOptions.logLevel]] are sent to each of these `LogSink`s.
-   * If you would still like to logs to go to the console, include
-   * [[consoleLogSink]] in your array.
+   * If you would still like logs to go to the console, include
+   * [[consoleLogSink]] in the array.
    *
    * ```ts
    * logSinks: [consoleLogSink, myCloudLogSink],
--- a/src/replicache-options.ts
+++ b/src/replicache-options.ts
@@ -1,4 +1,4 @@
-import type {LogLevel} from '@rocicorp/logger';
+import type {LogLevel, LogSink} from '@rocicorp/logger';
 import type {Pusher} from './pusher';
 import type {Puller} from './puller';
 import type {MutatorDefs, RequestOptions} from './replicache';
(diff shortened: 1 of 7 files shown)


src/replicache-options.ts
```diff
@@ -81,9 +81,18 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
...
    */
   logLevel?: LogLevel;
 
+  /**
+   * Enables custom handling of logs. By default logs are logged to the console.
+   * If you would like logs to be sent somewhere else (e.g. to a cloud logging
+   * service like DataDog) you can provide a custom `LogSink`.  Logs at or above
+   * [[ReplicacheOptions.logLevel]] are sent to this `LogSink`.
+   */
+  logSink?: LogSink;
```
aboodman: The thing that irritates me is that I think everyone will _always_ want the logs to go to console, regardless of whether they also go to DataDog. I can't think of a reason you would _not_ want them to go to console.

OTOH, providing a sink API and *not* providing a way to get rid of the console seems like a weird API.

What if instead we called this `logSinks` and it was an array of `LogSink` and by default it was the console logger. But people could `push()` on the datadog logger too if they wanted (which would be the typical usage of this feature).

aboodman: @arv

src/replicache.test.ts
```diff
@@ -1408,6 +1409,54 @@ test('logLevel', async () => {
...
+  let logCounts: Record<LogLevel, number> = initLogCounts();
+  const resetLogCounts = () => {
+    logCounts = initLogCounts();
+    infoStub.reset();
+    debugStub.reset();
+  };
+
+  const logSink = {
+    log: (level: LogLevel, ..._args: unknown[]) => {
+      logCounts[level] = logCounts[level] + 1;
```
arv: ++

src/replicache-options.ts
arv: I don't know. If I'm deploying gmail I'd be more than happy to not log to console by default. I would want more control.

I think, the idea to use an array is a bit too clever... and also, I don't like the mutability of options.

The `LogSink` is such a small interface that I think it is cleaner to let people do their own `TeeLogger`.

arv (approved):
You need to export LogSink from mod.ts too

aboodman:
> If I'm deploying gmail I'd be more than happy to not log to console by default.

Why? It seems useful to have the logs there, so that e.g., advanced users can report errors easily.

aboodman:
Certainly for my own apps I want the console logger and data dog. And having to implement / figure out TeeLogger feels like a lot of work for what to me seems like the common case.

src/replicache-options.ts
grgbkr: I like the idea of making it very easy to do console logging, plus cloud service logging.  I don't like the idea of them mutating a logSink array with push.  ReplicacheOptions is just an interface right now, its not a class you instantiate and then can mutate.  

I think a reasonable approach is to make this option
`logSinks?: LogSink[]`
to get data dog and console logging they can just do
```
import { consoleLogSink } from "replicache";

const options = {
  logSinks: [consoleLogSink, myCustomLogSink],
   ...
}
```

and then we can auto wrap these in a TeeLogger under the covers.

src/replicache.test.ts
grgbkr: done

src/replicache-options.ts
grgbkr: Updated to logSinks array.  what do you all think?

grgbkr: I also updated reflect to take an array of LogSinks for consistency https://github.com/rocicorp/reflect/pull/104

grgbkr:
> You need to export LogSink from mod.ts too

done

src/replicache-options.ts
```diff
@@ -81,9 +81,26 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
...
+   *
+   * By default logs are logged to the console.  If you would like logs to be
+   * sent elsewhere (e.g. to a cloud logging service like DataDog) you can
+   * provide an array of `LogSink`s.  Logs at or above
+   * [[ReplicacheOptions.logLevel]] are sent to each of these `LogSink`s.
+   * If you would still like logs to go to the console, include
+   * [[consoleLogSink]] in the array.
+   *
+   * ```ts
+   * logSinks: [consoleLogSink, myCloudLogSink],
```
arv, line 99: We should verify that the JSDocs for consoleLogSink is showing up in the generated docs

arv (approved):

arv:
LGTM

src/replicache-options.ts
grgbkr: good call it does not. I removed the linking.

grgbkr: actually I was able to get the linking to work by changing the docusaurus-plugin-typedoc config to `excludeExternals: false`, which seems like the right thing since now it includes docs for `LogSink`, `LogLevel`, and `consoleLogSink`.

grgbkr: Left is `excludeExternals: true` and right is `excludeExternals: false`.
image

image

arv: 👍

PR #914, opened 2022-04-05
refactor: Split out a ReplicacheInternal that can have internal options.

grgbkr:


grgbkr:
A few problems with this:
1. To proxy the properties (e.g. Replicache.pullURL) I have to add getters and setters.   This doesn't really change the api but it make tsdoc produce really ugly docs
property
image (1)

vs getter and setter
image (2)

2.  ReplicacheInternal still has to be exposed for Reflect.  Theoretically we could create two different npm packages, public one that exposes Replicache and private one that expose ReplicacheInternal... but I don't know how to do this in the same directory.

PR #915, opened 2022-04-05
feature: Add some internal hidden options needed for reflect

grgbkr:
Adds
1. disableLicensing
2. disableMutationRecovery

### DIFF
--- a/src/replicache-mutation-recovery.test.ts
+++ b/src/replicache-mutation-recovery.test.ts
@@ -26,6 +26,7 @@ import sinon from 'sinon';
 // @ts-expect-error
 import fetchMock from 'fetch-mock/esm/client';
 import {initClientWithClientID} from './persist/clients-test-helpers.js';
+import type {ReplicacheOptions} from './replicache-options';
 
 initReplicacheTesting();
 
@@ -1149,6 +1150,19 @@ test('mutation recovery returns early without running if push is disabled', asyn
     {
       pullURL: 'https://diff.com/pull',
     },
+  );
+  expect(rep.recoverMutationsSpy.callCount).to.equal(1);
+  expect(await rep.recoverMutationsSpy.firstCall.returnValue).to.equal(false);
+  expect(await rep.recoverMutations()).to.equal(false);
+});
+
+test('disableMutationRecovery hidden option', async () => {
+  const rep = await replicacheForTestingNoDefaultURLs(
+    'mutation-recovery-startup',
+    {
+      pullURL: 'https://diff.com/pull',
+      disableMutationRecovery: true,
+    } as unknown as ReplicacheOptions<Record<string, never>>,
   );
   expect(rep.recoverMutationsSpy.callCount).to.equal(1);
   expect(await rep.recoverMutationsSpy.firstCall.returnValue).to.equal(false);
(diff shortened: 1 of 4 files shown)


src/replicache.ts
```diff
@@ -279,6 +279,9 @@ export class Replicache<MD extends MutatorDefs = {}> {
   private _persistIsScheduled = false;
   private _recoveringMutations = false;
 
+  private readonly _disableLicensing: boolean = false;
```
phritz, line 282: I don't know how your brains work but for mine "enableFoo" is so much easier to keep in mind than "disableFoo" which if on/true has the effect of turning something off.&lt;/shed&gt;

phritz (approved):
LGTM nice and simple!

src/replicache.ts
grgbkr: I don't feel strongly at all about this.  

My choice of disable vs enable is based on the default.  If the default behavior is enabled, then the option is to disable it. 

For me
```
new Replicache({
  disableLicensing: true
});
```

scans faster than

```
new Replicache({
  enableLicensing: false 
});
```

aboodman: I don't care if it's internal API.

arv: I'm also in favor of enable over disable in general.

Having default be true is not an issue to me

src/replicache.ts
```diff
@@ -356,6 +359,17 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+      disableLicensing?: boolean;
+      disableMutationRecovery?: boolean;
+    };
+    if (internalOptions.disableLicensing) {
+      this._disableLicensing = true;
+    }
+    if (internalOptions.disableMutationRecovery) {
+      this._disableMutationRecovery = true;
+    }
+
```
arv, line 372: I would do something more along the lines of:

```ts
type InternalOptions = {
  disableLicensing?: boolean;
  disableMutationRecovery?: boolean;
};
const {
  disableLicensing = false,
  disableMutationRecovery = false,
} = options as InternalOptions;
this._disableLicensing = disableLicensing;
this._disableMutationRecovery = disableMutationRecovery;
```

And then remove the initializer in the class field declarations.

src/replicache-mutation-recovery.test.ts
```diff
@@ -1155,6 +1156,19 @@ test('mutation recovery returns early without running if push is disabled', asyn
   expect(await rep.recoverMutations()).to.equal(false);
 });
 
+test('disableMutationRecovery hidden option', async () => {
+  const rep = await replicacheForTestingNoDefaultURLs(
+    'mutation-recovery-startup',
+    {
+      pullURL: 'https://diff.com/pull',
+      disableMutationRecovery: true,
+    } as unknown as ReplicacheOptions<Record<string, never>>,
```
arv, line 1165: Maybe change `replicacheForTestingNoDefaultURLs` to allow the internal options so the tests do not all have to do this casting dance.

arv (commented):
Concept LGTM. Just a few nits

src/replicache.ts
grgbkr: done

src/replicache-mutation-recovery.test.ts
grgbkr: done

grgbkr:
@phritz @arv thanks for the reviews, some follow up improvements based on your comments in https://github.com/rocicorp/replicache/pull/927

PR #927, opened 2022-04-07
refactor: Small improvements to internal options

grgbkr:
1. Add a ReplicacheInternalOptions type def.
2. Have replicacheForTesting accept internal options so tests dont have to cast.
3. Switch options from disable to enable.

### DIFF
--- a/src/replicache.test.ts
+++ b/src/replicache.test.ts
@@ -2102,10 +2102,10 @@ async function licenseKeyCheckTest(tc: LicenseKeyCheckTestCase) {
   const {enableLicensing = true} = tc;
 
   const options = enableLicensing
-    ? ({
+    ? {
         licenseKey: tc.licenseKey,
         disableLicensing: true,
-      } as unknown as ReplicacheOptions<Record<string, never>>)
+      }
     : {licenseKey: tc.licenseKey};
 
   const rep = await replicacheForTesting(name, options);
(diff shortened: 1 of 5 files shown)


src/replicache.test.ts
```diff
@@ -2099,11 +2099,13 @@ async function licenseKeyCheckTest(tc: LicenseKeyCheckTestCase) {
...
   }
 
-  const options = tc.disableLicensing
-    ? ({
+  const {enableLicensing = true} = tc;
+
+  const options = enableLicensing
+    ? {
         licenseKey: tc.licenseKey,
         disableLicensing: true,
```
phritz: I'm unclear on why we would have an enableLicensing flag that sets a disableLicensing flag? I don't have to be clear on it tho, and feel bad for asking more questions about a boolean :)

grgbkr: because its a bug :)

grgbkr: it is very very weird that typescript didnt catch this...

grgbkr: also odd was that the test didn't fail.  this was because of a subtle bug in the existing test, because of a gotcha with fetchMock.  fetchMock only records call it was instructed to expect.  

So, the following always passes when expectFetchCalled is false, whether or not statusUrlMatcher is fetched.
```
  if (tc.expectFetchCalled) {
    fetchMock.postOnce(statusUrlMatcher, tc.mockFetchParams);
  }
  expect(fetchMock.called(statusUrlMatcher)).to.equal(tc.expectFetchCalled); `
```

To fix this I added a 
`fetchMock.catch();`
(which is like a default expect).

arv (commented):
LGTM

PR #929, opened 2022-04-08
Aa/scan result

aboodman:
This gets part of the way to making it possible to implement ScanResult on the server, but it doesn't quite work. See also https://github.com/rocicorp/replicache-todo/pull/4 where I tried to integrate this.

### DIFF
--- a/src/transactions.ts
+++ b/src/transactions.ts
@@ -8,7 +8,7 @@ import {
   ScanOptions,
   toDbScanOptions,
 } from './scan-options';
-import {ScanResultImpl} from './scan-iterator';
+import {ScanResult} from './scan-iterator';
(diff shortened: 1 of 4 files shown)


src/transactions.ts
```diff
@@ -50,7 +50,7 @@ export interface ReadTransaction {
    * If the [[ScanResult]] is used after the `ReadTransaction` has been closed it
    * will throw a [[TransactionClosedError]].
    */
-  scan(): ScanResultImpl<string, ReadonlyJSONValue>;
+  scan(): ScanResult<string, ReadonlyJSONValue>;
```
arv, line 53: This is the bug. The interface should not have the Impl as the return type.

PR #932, opened 2022-04-08
refactor: Convert the key for btree index iterator early

arv:
Instead of converting the key from third party BTree iterators to our
encoded string, convert our encoded string to to an entry. Then let the
main scan loop work with IndexKey as needed.

The benefit is that for external iterators we do not have to go from
IndexKey to string and back to IndexKey.

### DIFF
--- a/src/scan-iterator.ts
+++ b/src/scan-iterator.ts
@@ -208,10 +208,12 @@ async function* scanIterator<Options extends ScanOptions, V>(
     if (exclusive) {
       exclusive = true;
       if (isIndexScan) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         if (shouldSkipIndexScan(key as IndexKey, options.start!.key)) {
           continue;
         }
       } else {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         if (shouldSkipNonIndexScan(key as string, options.start!.key)) {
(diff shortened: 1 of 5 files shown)


src/scan-iterator.ts
```diff
@@ -281,41 +289,39 @@ export function makeScanResult<Options extends ScanOptions>(
...
+    ) as AsyncIter;
+    return new ScanResultImpl(
+      iter,
+      options,
+      {closed: false, shouldDeepClone: false},
+      _ => {
+        // noop
+      },
     );
-    internalIter = internalIndexScanIterator(iter);
```
arv, line 292: Here is the important change

src/transactions.ts
```diff
@@ -379,9 +391,11 @@ function makeScanResultFromScanIteratorInternal<
 async function* getScanIteratorForIndexMap(
   dbRead: db.Read,
   options: ScanIndexOptions,
-) {
+): AsyncIterable<IndexKeyEntry> {
   const map = await dbRead.getMapForIndex(options.indexName);
-  yield* map.scan(fromKeyForIndexScanInternal(options));
+  for await (const entry of map.scan(fromKeyForIndexScanInternal(options))) {
```
arv, line 396: And here is where we convert the internal iterator from encoded strings to IndexKey

PR #937, opened 2022-04-12
make cli actually work

phritz:
No idea how this slipped pass or even how it compiled -- vscode shows me an error that the parameter is missing but esbuild happily builds it.


arv:
esbuild does no type checking. It just strips the types.

It seems like we are not type checking the cli tool.

PR #942, opened 2022-04-12
fix: Prevent too many modules error in chrome

arv:
Chromium has a bug where it fails to load if there are too many modules
in the dependency tree. We were depending on lodash-es which has hundreds
of modules which all tried to get loaded in Chrome.

Change to only load the sub modules of lodash-es.


aboodman:
Nice. I wonder if it will effect tests vis gc pressure.

arv:
The sad/funny thing is that we ran into this exact issue with Chrome at Quip when we migrated to ES modules (I tried to find the chrome bug but no luck)

arv:
I guess it didn't work ;(

PR #950, opened 2022-04-14
chore: Make perf test use out/replicache

arv:
This is so that we are testing something closer to what we are building.

### DIFF
--- a/perf/replicache.ts
+++ b/perf/replicache.ts
@@ -8,7 +8,7 @@ import {
   Replicache,
   ReplicacheOptions,
   WriteTransaction,
-} from '../src/mod';
+} from '../out/replicache';
 import {jsonArrayTestData, TestDataObject, jsonObjectTestData} from './data';
 import type {Bencher, Benchmark} from './perf';
 import * as kv from '../src/kv/mod';
@@ -62,12 +62,16 @@ export function benchmarkPopulate(opts: {
   };
 }
 
+const persistSymbol = Symbol();
+
 class ReplicacheWithPersist<MD extends MutatorDefs> extends Replicache {
+  [persistSymbol]: () => void;
   constructor(options: ReplicacheOptions<MD>) {
-    super(options);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    super({...options, exposePersistMethodAs: persistSymbol} as any);
   }
   async persist(): Promise<void> {
-    return this._persist();
+    return this[persistSymbol]();
   }
 }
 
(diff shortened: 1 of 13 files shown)


src/replicache-options.ts
```diff
@@ -224,4 +224,9 @@ export type ReplicacheInternalOptions = {
    * Defaults to true.
    */
   enableMutationRecovery?: boolean;
+
+  /**
+   * Allows exposing _persist to perf test
+   */
+  exposePersistMethodAs?: symbol;
```
arv, line 231: @grgbkr This is another new "hack" to expose internals.

When we minimize we rename all properties that start with `_` so `_persist` gets renamed.

Instead we expose the method on this using a symbol which the subclass passes in.

rollup.config.js
```diff
@@ -6,13 +6,16 @@ import dts from 'rollup-plugin-dts';
 // We use esbuild for building the actual code.
 
 export default {
-  input: 'out/.dts/src/mod.d.ts',
+  input: 'out/.dts/mod.d.ts',
```
arv, line 9: argh...

The output uses the shortest common sub path and now that perf/ is no longer part of the project that changed.

src/replicache-options.ts
grgbkr: I don't love it.  I don't hate it.  I don't have a better suggestion.

perf/replicache.ts
```diff
@@ -62,12 +62,18 @@ export function benchmarkPopulate(opts: {
...
   constructor(options: ReplicacheOptions<MD>) {
-    super(options);
+    super({
+      ...options,
+      exposePersistMethodAs: persistSymbol,
+    } as ReplicacheOptions<MD>);
   }
   async persist(): Promise<void> {
-    return this._persist();
+    return this[persistSymbol]();
```
grgbkr, line 76: surprised this type checks.

tool/build.mjs
```diff
@@ -49,27 +52,18 @@ async function buildCLI() {
...
-    outfile: 'perf/index.js',
-    entryPoints: ['perf/index.ts'],
-    define: {
-      'process.env.NODE_ENV': '"production"',
-    },
-    sourcemap: true,
-    format: 'esm',
-    minify: true,
-  });
+  // Same as what wev build for the npm package but we turn on sourcemaps.
```
grgbkr, line 55: I think we should also always publish a private package of replicache with source maps.  It's hard to debug issues in our sample apps (that we suspect may be a replicache bug) without source maps.

src/replicache-options.ts
grgbkr: Something that may type check a bit better, and also can more easily expose additional internal apis.
```
interface ReplicacheInternalOptions {
  receiveInternalAPI?: (internalAPI: ReplicacheInternalAPI) => void;
}

interface ReplicacheInternalAPI {
  persist(): Promise<void>;
}
```

grgbkr (commented):
LGTM

tool/build.mjs
arv: That is a good idea... Or we could publish a private package that is not minimized? But sourcemaps might be a better start.

PR #951, opened 2022-04-14
refactor: Add AbortSignal to sleep

arv:
I need it in the near future

### DIFF
--- a/src/sleep.ts
+++ b/src/sleep.ts
@@ -1,14 +1,29 @@
+import {AbortError} from './abort-error';
+
 /**
  * Creates a promise that resolves after [[ms]] milliseconds. Note that if you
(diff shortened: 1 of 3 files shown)


src/sleep.ts
```diff
@@ -1,14 +1,29 @@
...
+    const id = setTimeout(() => {
+      resolve();
+    }, ms);
+
+    if (signal) {
+      signal.addEventListener('abort', () => {
+        if (id) {
+          clearTimeout(id);
+        }
+        reject(new AbortError('Aborted'));
```
arv, line 25: I initially wanted DOMException to match fetch, but it is not available in NodeJS yet. Not sure about CloudFlare at this point.

grgbkr:
Nice.  LGTM.

PR #952, opened 2022-04-14
chore; Add lastOpenedTimestampMS to IndexedDBDatabase

arv:
When the IDB is added to the meta table we note the time stamp. This is
going to be used later when we GC IDB instances.


grgbkr:
LGTM

PR #955, opened 2022-04-15
feat: Collect old IDB databases

### LINKED ISSUE #803: Garbage collect old IDB instances
When we do `new Replicache({name: 'x'})` we create an IDB instance that maps to `'x'`. If the app is of the nature that it generates unique names (for example for different rooms) we end up creating a lot of IDB instances that might never be accessed again.

We should delete old IDB databases that are older than 90 days (for example) to reduce the amount of disk storage the Replicache IDB databases take up.

arv:
We now collect IDB databases that have not been opened for 3 months.

We check 5 minutes after startup as well as every 12 hours. In theory we
should be able to check much less frequently. We could compute when the
oldest one would expire and not check until then...

Fixes rocicorp/replicache#803, rocicorp/mono#121

### DIFF
--- a/src/persist/collect-idb-databases.test.ts
+++ b/src/persist/collect-idb-databases.test.ts
@@ -1,0 +1,236 @@
+import {expect} from '@esm-bundle/chai';
+import {fakeHash} from '../hash.js';
+import {TestMemStore} from '../kv/test-mem-store';
+import {makeClientMap, setClients} from './clients-test-helpers';
+import {
+  IDBDatabasesStore,
+  IndexedDBDatabase,
+  IndexedDBName,
+} from './idb-databases-store';
+import {collectIDBDatabases} from './collect-idb-databases';
+import * as dag from '../dag/mod';
+import type {ClientMap} from './clients.js';
+import {assertNotUndefined} from '../asserts.js';
+import {SinonFakeTimers, useFakeTimers} from 'sinon';
+import {REPLICACHE_FORMAT_VERSION} from '../replicache.js';
+
+suite('collectIDBDatabases', async () => {
+  let clock: SinonFakeTimers;
+
+  setup(() => {
+    clock = useFakeTimers(0);
+  });
+
+  teardown(() => {
+    clock.restore();
+  });
+
+  type Entries = [IndexedDBDatabase, ClientMap][];
+
+  const makeIndexedDBDatabase = (
+    name: string,
+    lastOpenedTimestampMS = Date.now(),
+    replicacheFormatVersion = REPLICACHE_FORMAT_VERSION,
+    schemaVersion = 'schemaVersion-' + name,
+    replicacheName = 'replicacheName-' + name,
+  ): IndexedDBDatabase => ({
+    name,
+    replicacheFormatVersion,
+    schemaVersion,
+    replicacheName,
+    lastOpenedTimestampMS,
+  });
+
+  const t = (
+    name: string,
+    entries: Entries,
+    now: number,
+    expectedDatabases: string[],
+  ) => {
+    for (const legacy of [false, true]) {
+      test(name + ' > time ' + now + (legacy ? ' > legacy' : ''), async () => {
+        const store = new IDBDatabasesStore(_ => new TestMemStore());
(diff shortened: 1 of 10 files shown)


aboodman:
Thanks for getting rocicorp/mono#121 in here too.

- Wasn't there already some code for collecting old *clients* in the same idb instance? Should it be merged into this so all GC related code is shared?
- I think 1 month is a better value than 3 months. For users like Vercel these will pile up fast (all documents are their own  IDB).

Note: the client-collecting code I believe uses the value 1 week for the timeout, but I think it makes sense that clients have a shorter timeout than idbs, because clients are really sessions. If a session is older than say, a day, old chances of it getting revisited quickly go to zero.

src/persist/collect-idb-databases.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  // 0 is used in testing
+  if (db.lastOpenedTimestampMS !== undefined) {
+    return now - db.lastOpenedTimestampMS >= maxAge;
+  }
+
+  // For legacy databases we do not have a lastOpenedTimestampMS so we check the
+  // time stamps of the clients
+  const perdag = newDagStore(db.name);
```
grgbkr, line 144: Should we wrap the attempt to read legacy dbs in a try catch to be robust against current code having a conflict with legacy db format?

src/persist/collect-idb-databases.ts
```diff
@@ -0,0 +1,162 @@
...
+) {
+  try {
+    await sleep(COLLECT_DELAY, signal);
+
+    await collectIDBDatabases(
+      idbDatabasesStore,
+      signal,
+      Date.now(),
+      MAX_AGE,
+      defaultNewDagStore,
```
grgbkr, line 59: perhaps don't pass defaultNewDataStore since it is the default for this param?

src/persist/collect-idb-databases.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  const stopInterval = initBgIntervalProcess(
+    'CollectIDBDatabases',
+    async () => {
+      await collectIDBDatabases(
+        idbDatabasesStore,
+        signal,
+        Date.now(),
+        MAX_AGE,
+        undefined,
```
grgbkr, line 37: perhaps don't pass undefined since it is an optional param?

src/persist/collect-idb-databases.ts
```diff
@@ -0,0 +1,162 @@
...
+    }),
+  );
+
+  const idbRemovedNames: string[] = [];
+  const errors: unknown[] = [];
+  for (const result of dropStoreResults) {
+    if (result.status === 'fulfilled') {
+      idbRemovedNames.push(result.value);
+    } else {
+      errors.push(result.reason);
```
grgbkr, line 108: Is there a reasonable way for us to detect if the delete failed because the db no longer exists, so we can remove that db from the meta table?

src/persist/collect-idb-databases.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        }),
+      ],
+    ];
+    t('one idb, one client, format version too new', entries, 0, ['a']);
+    t('one idb, one client, format version too new', entries, 1000, ['a']);
+    t('one idb, one client, format version too new', entries, 2000, ['a']);
+  }
+
+  {
+    const entries: Entries = [
```
grgbkr, line 222: suggest adding a test for a old format version db with multiple clients.

grgbkr (commented):
LGTM

src/persist/collect-idb-databases.ts
arv: `indexedDB.deleteDatabase('no-such-db')` succeeds.

arv: The failure case I found is that the DB is in use.

arv:
> * Wasn't there already some code for collecting old _clients_ in the same idb instance? Should it be merged into this so all GC related code is shared?

The code that collects clients is very different. It rewrites the clients map and then the dag gc collects the chunks that are no longer reachable. This code is checking the last opened time and deletes the whole IDB database based on that. I don't see a reason to try share any code here.

PR #960, opened 2022-04-18
chore: Use extends instead of hack

arv:
I forgot to fix/remove this `@ts-expect-error` when I was working on
scan.


grgbkr:
LGTM

PR #969, opened 2022-04-18
wat

phritz:
This is a typescript thing I do not understand. The cli.ts file invokes `licensingCLI.main()` without the [required argument](https://github.com/rocicorp/licensing/blob/021cb42bdbd65c10b1d4ea0a609a56b8a3f394e9/src/cli/get-license.ts#L36). In VSCode that line is red showing me the expected error `Expected 1 arguments, but got 0. An argument for 'licenseServerURL' was not provided.`. However it compiles just fine and then fails at runtime when the missing argument is attempted to be used. To see this:
1. `rm $(which replicache)`
2. `rm out/cli.cjs`
3. `npm run build`
4. `npm link`
5. `npx replicache get-license`
6. enter data at the prompts 
7. notice that in the final step when it goes to use the missing argument it fails:
```
Fetching License Key...
Sorry, there was an error. Please file a bug at https://github.com/rocicorp/replicache/issues/new with the following information:
TypeError [ERR_INVALID_URL]: Invalid URL
```
The "invalid url" is `undefined`.

### DIFF
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -15,4 +15,4 @@ if (process.argv.length < 3 || process.argv[2] !== 'get-license') {
 
 // TODO(phritz): add staging flag
 
-licensingCLI.main(PROD_LICENSE_SERVER_URL);
+licensingCLI.main();


phritz:
Interesting that CI catches the problem (this time, it didn't a few weeks ago?!) bc it runs `check-types`. But shouldn't build check the types?!

aboodman:
I think this is just a code factoring / terminology thing.

The `build` step doesn't include type checking. That might seem crazy, but in js land everything is configurable, swappable, customizable, etc. And @arv has opted to use `esbuild` for compilation for ... reasons ... instead of tsc (I think partly because it's super fast, but probably other reasons too), but esbuild doesn't do type checking. So we use `tsc` to do that. But the `build` script doesn't seem to include that step:

https://github.com/rocicorp/replicache/blob/main/tool/build.mjs

aboodman:
(I agree that intuitively, I would expect `npm run build` to do type checking. Not sure if there's a reason it doesn't currently, or just accidental).

phritz:
Seems weird to go out of our way to use a compile-time type language and then not use it at compile time!

phritz:
@arv anything more that I should understand here?

arv:
I think we should change our build script to also run the type checking.

phritz:
> I think we should change our build script to also run the type checking.

do you mean just having a rule like:
```
"prebuild": "npm run check-types"
```
or do you mean monkeying with the esbuild parameters in https://github.com/rocicorp/replicache/blob/main/tool/build.mjs?

arv:
`prebuild` works for me but there might be some reconciliation that should be done. I can take over if you want?

phritz:
> prebuild works for me but there might be some reconciliation that should be done. I can take over if you want?

yes please and thanks :)

Issue #839, opened 2022-02-13
`watch()`


aboodman:
One of our users maintains state outside of Replicache (in a separate indexeddb instance) similar to #497 proposes. This is done to implement full text search. However it's not possible to maintain this state efficiently because we hide the raw change notification (key-added, key-modified, key-removed) from users and only make the results of subscribe available.

aboodman:
See https://discord.com/channels/830183651022471199/830183651022471202/941897663077576755

arv:
Seems easy enough to implement. The only question is if we want to expose a single callback that takes all the changes or if we want to call the callback mutliple times with a single change each time. Gut feeling is to call with all the changes as an array.

aboodman:
I guess a single callback would be best for performance reasons. Maybe resurrect `onChange` and have it receive a `Patch` ?

aboodman:
I think it is going to be important for correctness reasons for this callback to receive as a parameter the previous rootHash and the new rootHash. Users that are trying to track Replicache state in some other store (e.g., another idb) are going to need to store the newRootHash in their store (transactionally) then compare the incoming (prevRootHash) to the one they have stored to ensure they are applying the change against the correct state.

aboodman:
So conceptually, what this API looks like to me is:

```ts
function onChange(from: Hash, to: Hash, patch: Patch) {

}
```

I am not sure if this API should be synchronous or asynchronous with respect to the causing Replicache transaction. I can imagine pros and cons both ways.

tmcw:
Frankly I'd be interested in this as an alternative to `scan()`. If I'm running a single scan on startup then updating that datastructure in response to change notifications, it seems feasible to get performance relative to the amount of data changing, rather than the total amount of data. As it stands, large datasets (40k features, about 250 bytes each) don't scan quickly enough to be updated interactively.

aboodman:
@tmcw's use case is similar except that caller is really only interested in a subset of the keyspace.

Also, I think that since there will eventually need to be a variant of this API that runs inside the transaction, so I've renamed it to make usage more clear (analagous to `onbeforeulnoad` vs `onunload`).

Updated API strawperson:

```ts
class Replicache {
  onAfterCommit(prefix: string, handler AfterCommitHandler): Unwatcher;
  // future: onBeforeCommit(prefix: string, handler: BeforeCommitHandler): Unwatcher;
}

// from and to can be used to keep separate storage in sync
[...]

PR #974, opened 2022-04-20
refactor: Use DiffResult all the way

arv:
Instead of having a `string[]` with the changed keys we now use
`DiffResult<ReadonlyJSONValue>[]` all the way.

TODO: This does not rename things yet.

Towards #839


src/sync/pull.ts
```diff
@@ -245,10 +246,12 @@ export type ReplayMutation = {
 // "" is used for the primary index.
 export type ChangedKeysMap = Map<string, string[]>;
 
+export type ChangedDiffsMap = Map<string, DiffResult<ReadonlyJSONValue>[]>;
```
grgbkr, line 249: ChangedDiffs seems redundant.  Perhaps DiffsMap, or DiffResultsMap.

src/db/write.ts
```diff
@@ -264,30 +265,28 @@ export class Write extends Read {
...
     const valueHash = await this.map.flush();
-    let valueChangedKeys: string[] = [];
-    if (generateChangedKeys && this._basis) {
+    let valueDiff: DiffResult<ReadonlyJSONValue>[] = [];
+    if (generateDiffs && this._basis) {
       const basisMap = new BTreeRead(this._dagWrite, this._basis.valueHash);
-      valueChangedKeys = await asyncIterableToArray(
-        this.map.diffKeys(basisMap),
-      );
+      valueDiff = await asyncIterableToArray(this.map.diff(basisMap));
```
grgbkr, line 280: should this import diff from diff.ts and use that?

src/db/write.ts
```diff
@@ -296,20 +295,24 @@ export class Write extends Read {
...
-              asyncIterableToArray(map.diffKeys(basisMap)),
-            );
-          }
-          return asyncIterableToArray(map.keys());
-        },
-      );
+      const indexDiffResult = await index.withMap(this._dagWrite, async map => {
+        if (basisIndex) {
+          return basisIndex.withMap(this._dagWrite, basisMap =>
+            asyncIterableToArray(map.diff(basisMap)),
```
grgbkr, line 301: should this use diff from diff.ts?

src/sync/pull.ts
```diff
@@ -405,18 +408,26 @@ function assertResult(v: any): asserts v is Result {
 
   assertHTTPRequestInfo(v.httpRequestInfo);
 }
+
 async function addChangedKeysForIndexes(
```
grgbkr, line 412: addChangedKeysForIndexes  => addDiffsForIndexes

src/sync/pull.ts
```diff
@@ -427,18 +438,18 @@ async function addChangedKeysForIndexes(
       const newIndex = newIndexes.get(oldIndexName);
       if (newIndex !== undefined) {
         const changedKeys = await newIndex.withMap(read, async newMap => {
```
grgbkr, line 440: changedKeys => diffs

src/sync/pull.ts
```diff
@@ -427,18 +438,18 @@ async function addChangedKeysForIndexes(
...
         newIndexes.delete(oldIndexName);
         if (changedKeys.length > 0) {
-          changedKeysMap.set(oldIndexName, changedKeys);
+          diffsMap.set(oldIndexName, changedKeys);
         }
       } else {
-        // old index name is not in the new indexes. All keys changed!
-        const changedKeys = await allKeys(oldMap);
+        // old index name is not in the new indexes. All entries changed!
+        const changedKeys = await diffForAll(oldMap);
```
grgbkr, line 450: changedKeys => diffs

src/sync/pull.ts
```diff
@@ -447,9 +458,9 @@ async function addChangedKeysForIndexes(
   for (const [newIndexName, newIndex] of newIndexes) {
     // new index name is not in the old indexes. All keys changed!
     await newIndex.withMap(read, async newMap => {
-      const changedKeys = await allKeys(newMap);
+      const changedKeys = await diffForAll(newMap);
```
grgbkr, line 461: changedKeys => diffs

src/sync/pull.ts
```diff
@@ -447,9 +458,9 @@ async function addChangedKeysForIndexes(
...
       if (changedKeys.length > 0) {
-        changedKeysMap.set(newIndexName, await allKeys(newMap));
+        diffsMap.set(newIndexName, await diffForAll(newMap));
```
grgbkr, line 463: We don't need to compute diffForAll again, we can just use the value computed above (currently named changedKeys, but should be named diffs).

src/sync/pull.ts
grgbkr, line 450: This needs to be all deleted, not all added, correct?  I think diffForAll need to take an op.

grgbkr: Seems we probably need a test for this.

grgbkr (approved):

src/db/write.ts
```diff
@@ -296,20 +295,24 @@ export class Write extends Read {
     for (const [name, index] of this.indexes) {
```
grgbkr, line 295: I think this is missing handling for indexes that are in basisIndexes but missing from this.indexes.

src/sync/pull.ts
arv: I see what you are saying now... Will look

src/db/write.ts
arv: Not sure. Filing an issue to not forget this

arv: You are correct. This is not dealing with that case. Good catch.

PR #976, opened 2022-04-20
feat: Add watch function

### LINKED ISSUE #839: `watch()`

arv:
Now you can get called after a commit is complete and the arguments
passed to the watch function includes a diff of the changes.

Closes #839


arv:
~~This is not the final API!!!~~

This is now the final API

src/btree/node.ts
```diff
@@ -26,6 +26,8 @@ export function getRefs(node: Node): ReadonlyArray<Hash> {
   return isInternalNode(node) ? node[NODE_ENTRIES].map(e => e[1]) : [];
 }
 
+export type Diff = readonly DiffOperation[];
```
arv, line 33: These needs JSDoc and get exposed in mod.ts

src/replicache-on-after-commit.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  expect(spy.lastCall.args).to.deep.equal([
+    [
+      {
+        op: 'del',
+        key: 'b',
+        oldValue: 2,
+      },
+    ],
+    't/000000000000000000000000000005',
+    't/000000000000000000000000000007',
```
arv: These temp hashes are a bit ugly.

One concern here is that once we get a persist, oldHash is not going to match newHash from the last onAfterCommit.

arv:
Now with final `watch` API

@aboodman Please review the API and the JSDoc

@grgbkr Please review logic in watcher.ts

src/watcher.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (potentialWatchers.size === 0) {
+    return foundMatchingWatchers;
+  }
+
+  for (const op of diff) {
+    for (const entry of potentialWatchers) {
+      if (op.key.startsWith(entry.prefix)) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        foundMatchingWatchers.get(entry)!.push(op);
+      } else if (op.key > entry.prefix) {
```
grgbkr: So the ops in diff are guaranteed to be in increasing key order?

grgbkr: If so we can optimize this a bit.  Instead of linearly scanning the whole diff, we can binary search for entry.prefix in diff, and then linear scan forward till op.key > entry.prefix.

arv: Yes. They are ordered.

I got lazy. Happy to do better in a follow up

grgbkr:
```
for (const entry of potentialWatchers) {
   const start = _.sortedIndexBy(diff, { key: entry.prefix }, (o) => o.key);
   for (let i = start; i < diff.length; i++) {
     const op = diff[i];
     if (op.key > entry.prefix) {
       break;
     }
     foundMatchingWatchers.get(entry)!.push(op);
   }
}

```

grgbkr: I think with this approach the overall structure of this function can be simplified as well.
```
export function findMatchingWatchers(
  diff: Diff,
  callbacks: Iterable<WatchCallbackEntry>,
): Map<WatchCallbackEntry, Diff> {
  const matches = new Map<WatchCallbackEntry, DiffOperation[]>();
  for (const entry of callbacks) {
     if (entry.prefix === '') {
       matches.set(entry, diff as DiffOperation[]);
     } else { 
       const start = _.sortedIndexBy(diff, { key: entry.prefix }, (o) => o.key);
       const matchingOps = [];
       for (let i = start; i < diff.length; i++) {
         const op = diff[i];
         if (op.key > entry.prefix) {
           break;
         }
         matchingOps.push(op);
       }
       matches.set(entry, matchingOps);
    }
  }
  return matches;
}
```

src/replicache-on-after-commit.test.ts
grgbkr: Seems temp hashes break the whole purpose of exposing these hashes in this api.

src/watcher.ts
arv: Thanks.

src/btree/node.ts
```diff
@@ -26,6 +26,19 @@ export function getRefs(node: Node): ReadonlyArray<Hash> {
   return isInternalNode(node) ? node[NODE_ENTRIES].map(e => e[1]) : [];
 }
 
+/**
+ * Describes the changes that happened to Replicache after a
+ * [[WriteTransaction]] was committed.
+ */
+export type Diff = readonly DiffOperation[];
```
aboodman, line 33: I thought we already had a `Patch` type lying around that has this same definition.

aboodman: Nevermind.

src/replicache.ts
```diff
@@ -1315,6 +1324,18 @@ export class Replicache<MD extends MutatorDefs = {}> {
     }
   }
 
+  /**
+   * Watches Replicache for changes.
```
aboodman, line 1328:
```suggestion
   * Watches Replicache keys for changes.
```

src/replicache.ts
```diff
@@ -1315,6 +1324,18 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+   *
+   * The `callback` gets called whenever the underlying data changes and the
+   * `key` changes matches the [[WatchOptions.prefix]] if present.
+   *
+   * This gets called after commit (a mutation or a rebase).
```
aboodman, line 1333:
```suggestion
   * This gets called after a write transaction has committed (either a mutation or a rebase) and the change is final.
```

src/watcher.ts
```diff
@@ -0,0 +1,108 @@
+import {assert} from './asserts.js';
+import type {Diff, DiffOperation} from './btree/node.js';
+import type {Hash} from './hash.js';
+import type {DiffsMap} from './sync/pull.js';
+
+/**
+ * Function that gets passed into [[Replicache.watch]] and gets called when the
+ * data in Replicache changes.
```
aboodman:
```suggestion
 * data in Replicache changes.
 *
 * `oldRootHash` and `newRootHash` are strings that represent the entire state of the client view before and after the change. These strings are internal to Replicache and have no semantics other than opaque identifiers. 
 *
 * Applications can use these values to synchronize external persistent state with Replicache, for example, in another `IndexedDB` database. By recording the `newRootHash` in the external database with each change, an application can know whether it is safe to just apply the patch to the external storage. If the `oldRootHash` does not match the last hash stored in the external database, then using the patch is not safe, and the application should rewrite the external state from scratch.
```

src/replicache.ts
```diff
@@ -1315,6 +1324,18 @@ export class Replicache<MD extends MutatorDefs = {}> {
...
+   */
```
aboodman, line 1334:
```suggestion
   * The returned function should be called to stop watching.
   */
```

aboodman (approved):

PR #980, opened 2022-04-21
feat: Add support for pending to makeScanResult

### LINKED ISSUE #970: `makeScanResult()` should take a `pending: AsyncIterator` of key/value pairs
Every implementation is going to need this and the implementation is subtle:

https://github.com/rocicorp/replicache-todo/pull/4/files#diff-576e83834357ff3a413e9d71a569e036314294a31aa5ed8bde4d368dfaa1289b

arv:
Closes #970


arv:
I think my initial reaction still stands:

Putting this into `makeScanResult` makes an already hard to understand API even harder to understand.

Having a `mergeAsyncIterables` function makes sense and it is a bit strange to bundle this with replicache.

arv:
Changed to draft to ensure not landed. This needs unit tests and JSDoc and maybe even refactoring of the zillion types that this introduces.

src/merge-async-iterables.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Merges an iterable on top of another iterable.
+ *
+ * The second iterable can support the concept of delete entries (tombstones)
+ * and if the value of the second iterable is a delete value then the value is
+ * not included in the resulting iterable.
+ */
+export async function* mergeAsyncIterables<A, B>(
```
aboodman, line 8: I see what you're saying. If you prefer to expose this and have callers use it that would be fine with me.

